PR #2, opened 2016-02-12
fix deprecated akka shutdown

### LINKED ISSUE #1: Improve Akka Shutdown
Technical debt from the Akka upgrade. Fix the Akka shutdown logic in Main. It looks like this will also affect BaseActorTest.

BrandonArp:
Fixes #1

### DIFF
--- a/src/main/java/com/arpnetworking/metrics/mad/Main.java
+++ b/src/main/java/com/arpnetworking/metrics/mad/Main.java
@@ -17,6 +17,7 @@ package com.arpnetworking.metrics.mad;
 
 import akka.actor.ActorSystem;
 import akka.actor.Props;
+import akka.actor.Terminated;
 import akka.http.javadsl.Http;
 import akka.http.javadsl.IncomingConnection;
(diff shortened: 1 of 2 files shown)


src/main/java/com/arpnetworking/metrics/mad/Main.java
```diff
@@ -264,14 +267,19 @@ private void shutdownActors() {
...
-    @SuppressWarnings("deprecation")
     private void shutdownAkka() {
         LOGGER.info().setMessage("Stopping akka").log();
 
         if (_actorSystem != null) {
-            _actorSystem.shutdown();
-            _actorSystem.awaitTermination();
+            Future<Terminated> terminate = _actorSystem.terminate();
+            try {
+                Await.result(terminate, Duration.create(5, TimeUnit.MINUTES));
```
vjkoskela: 5 minutes? This seems like a bad idea.

BrandonArp: Why?  Before it was waiting forever.  At least now it's bounded.

PR #6, opened 2016-03-31
Performance Fixes Round One (Redux)

vjkoskela:
Fixed the performance tests and separated test initialization from the actual test to provide more accurate cpu profiling. Upgraded dependenencies and migrated to build with constructor reference instead of reflection to improve performance. Still more work to be done here but this was a start.


BrandonArp:
Already re-applied.

PR #7, opened 2016-04-06
fix the logging shutdown hook

BrandonArp:


src/main/java/com/arpnetworking/metrics/mad/Main.java
```diff
@@ -80,6 +81,8 @@
      * @param args the command line arguments
      */
     public static void main(final String[] args) {
+        final AtomicReference<DynamicConfiguration> configRef = new AtomicReference<>();
+        final AtomicReference<Configurator<?, ?>> configuratorRef = new AtomicReference<>();
         LOGGER.info().setMessage("Launching mad").log();
```
vjkoskela: Move these below the exception handler setup.

src/main/java/com/arpnetworking/metrics/mad/Main.java
```diff
@@ -96,6 +99,25 @@ public static void main(final String[] args) {
                         .log()
         );
 
+        Runtime.getRuntime().addShutdownHook(
+                new Thread(
+                        () -> {
+                            LOGGER.info().setMessage("Stopping mad").log();
+                            final DynamicConfiguration configuration = configRef.get();
+                            if (configuration != null) {
```
vjkoskela: getAndSet

src/main/java/com/arpnetworking/metrics/mad/Main.java
```diff
@@ -96,6 +99,25 @@ public static void main(final String[] args) {
...
+                                configuration.shutdown();
+                            }
+                            final Configurator<?, ?> configurator = configuratorRef.get();
```
vjkoskela: getAndSet

src/main/java/com/arpnetworking/metrics/mad/Main.java
```diff
@@ -122,21 +144,10 @@ public static void main(final String[] args) {
                                 .build())
                 .addListener(configurator)
                 .build();
+        configRef.set(configuration);
```
vjkoskela: Same comment as CAgg review about initialization.

PR #8, opened 2016-04-06
create a docker build

BrandonArp:


vjkoskela:
As discussed here are three work items to complete this change:

1) Logging at debug is way too verbose, please button it down a little more by default.
2) Update the README to split the installation section into docker and non-docker and describe the intent to use the mad image as a base image and give examples of how to override its configuration
3) Have maven build the docker image on release and push it up.

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM java:8-jre-alpine
+EXPOSE 7090
+RUN mkdir -p /opt/mad/logs
+RUN mkdir -p /opt/mad/config/pipelines
+ADD target/appassembler /opt/mad
```
vjkoskela: Not sure exactly how to do it, but the ideal here would be to do this in layers:

1) Add the bin directory.
2) Add the lib directory minus the mad jar.
3) Add the mad jar.
4) Add the configuration.

This will minimize the change in the docker image from one release to the next.

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+ADD config /opt/mad/config
+WORKDIR /opt/mad
+ENV PARAMS /opt/mad/config/config.json
+ENV JAVA_OPTS -Dlogback.configurationFile=/opt/mad/config/logback.xml
```
vjkoskela: I would recommend exposing a little more control here. One, expose a specific environment variable for the logging configuration file path. Probably also one for the configuration path (which we can eventually support a uri for). Also support additional jvm options to be passed in. This way you can override parts of the runtime configuration without having to redefine all of it.

config/config.json
```diff
@@ -0,0 +1,25 @@
+{
+  "monitoringCluster": "mad",
+  "logDirectory": "logs",
+  "pipelinesDirectory": "/opt/mad/config/pipelines",
+  "httpHost": "0.0.0.0",
+  "httpPort": 7090,
+  "httpHealthCheckPath": "/ping",
+  "httpStatusPath": "/status.json",
```
vjkoskela: Thoughts us just using /status? -> implication being for CAgg that we merge status.json and the akka status endpoint.

config/config.json
```diff
@@ -0,0 +1,25 @@
...
+  "jvmMetricsCollectionInterval": "PT.5S",
+  "akkaConfiguration": {
+    "akka": {
+      "loggers": [
+        "akka.event.slf4j.Slf4jLogger"
+      ],
+      "loglevel": "DEBUG",
```
vjkoskela: Remind me again how this log level relates to the logback configuration?

config/logback.xml
```diff
@@ -0,0 +1,47 @@
...
+        <rollingPolicy class="ch.qos.logback.core.rolling.TimeBasedRollingPolicy">
+            <timeBasedFileNamingAndTriggeringPolicy class="com.arpnetworking.logback.SizeAndRandomizedTimeBasedFNATP">
+                <maxOffsetInMillis>900000</maxOffsetInMillis>
+                <maxFileSize>100MB</maxFileSize>
+            </timeBasedFileNamingAndTriggeringPolicy>
+            <fileNamePattern>/opt/mad/logs/mad.%d{yyyy-MM-dd_HH}.%i.log.gz</fileNamePattern>
+            <maxHistory>4</maxHistory>
+            <cleanHistoryOnStart>true</cleanHistoryOnStart>
+        </rollingPolicy>
+        <encoder class="com.arpnetworking.logback.KeyValueEncoder">
```
vjkoskela: This does not provide the same level of serialization support as the StenoEncoder. If we are adopting logback-steno throughout these projects I would suggest we use StenoEncoder by default in the image.

Dockerfile
BrandonArp: Tried to make it a little better.

Dockerfile
BrandonArp: Done

config/config.json
BrandonArp: Changed to /status.  We need to do something about CAgg.

config/config.json
BrandonArp: This is the minimum level that using the Akka logging will send across to the logging actor.  If this is set above DEBUG, you will never get DEBUG messages sent to slf4j.  Make sense?

config/config.json
vjkoskela: Change not reflected here.

config/config.json
vjkoskela: Yeah, it just gets uncached fro time to time, me we touch it so infrequently. It makes you  want to use hocon for config (w. comments).

vjkoskela:
Looks good minus the missing /status change. Feel free to merge yourself tomorrow with that change.

PR #10, opened 2016-04-14
setup docker publication

BrandonArp:


.travis.yml
```diff
@@ -8,12 +8,30 @@ env:
...
 install: true
 before_script:
 - gpg --import arpnetworking.key
 script:
 - curl -s ${JDKW} | bash /dev/stdin ./mvnw clean install --settings settings.xml -P ci -Dgpg.skip=true -U
-- if [ \( -n "${TRAVIS_TAG}" \) -a "${TRAVIS_PULL_REQUEST}" = "false" ]; then curl -s ${JDKW} | bash /dev/stdin ./mvnw deploy --settings settings.xml -P ci -DskipAllVerification=true; fi;
+- if [ \( -n "${TRAVIS_TAG}" \) -a "${TRAVIS_PULL_REQUEST}" = "false" ]; then
+-   curl -s ${JDKW} | bash /dev/stdin ./mvnw deploy --settings settings.xml -P ci -DskipAllVerification=true
+-   if [ $? -eq 0]; then
```
vjkoskela: Not sure this is going to work nicely with the pipe above. I recall we had issues with this before. Even if you don't want to put all these commands in pom.xml I think they should be externalized and kicked-off from the pom (e.g. via profile) to avoid exactly this type of issue. For example:

```bash
TARGET="install"
PROFILES="ci,"
if [ \( -n "${TRAVIS_TAG}" \) -a "${TRAVIS_PULL_REQUEST}" = "false" ]; then
    TARGET="${TARGET}docker,"
    PROFILES="${PROFILES}
fi
curl -s ${JDKW} | bash /dev/stdin ./mvnw clean ${TARGET} --settings settings.xml -P ${PROFILES} -Dgpg.skip=true -U
```

Then the docker profile adds the creation and publication of a docker image. This can be a script execution for now, but it's at least insulated behind the same standard entry point if I want to build this locally instead of tee'd off from the travis specific entry point.

What's wrong with this one?

https://github.com/spotify/docker-maven-plugin

- Supports docker files
- Supports multiple tags

Looks like we should use an assembly to build the layout (https://maven.apache.org/plugins/maven-assembly-plugin/directory-mojo.html) and then either add it directly to the image or add from it. This combination would allow us to nicely layer base image, dependencies, config, and code separately in the right order.

vjkoskela:
#23 should cover this.

PR #11, opened 2016-04-15
add an rpm build

BrandonArp:


rpm/fs/etc/init.d/mad
```diff
@@ -0,0 +1,167 @@
...
+}
+
+start() {
+    [ -f $config ] || exit 6
+    echo -n $"Starting $prog: "
+    if is_running; then
+        warning "$name: already running"
+    else
+        cd "$dir"
+        su "$user" -s /bin/bash -c "nice $exec $config >> $dir/system.out 2>&1 & echo \$! > $pid_file"
```
vjkoskela: Does this need to be bash specifically instead of say sh; I'm thinking whether we then need to declare it as a dependency for the rpm -- and if so, is it necessary to be bash instead of sh?

rpm/fs/etc/init.d/mad
```diff
@@ -0,0 +1,167 @@
+#!/bin/sh
```
vjkoskela: I would hope that it's a reasonable assumption to have sh available on the system. Do rpms usually declare this dependency?

pom.xml
```diff
@@ -533,5 +534,82 @@
...
+                <scriptFile>rpm/scripts/post.sh</scriptFile>
+              </postinstallScriptlet>
+              <preremoveScriptlet>
+                <scriptFile>rpm/scripts/preun.sh</scriptFile>
+              </preremoveScriptlet>
+              <postremoveScriptlet>
+                <scriptFile>rpm/scripts/postun.sh</scriptFile>
+              </postremoveScriptlet>
+              <requires>
+                <require>java &gt;= 1.8.0</require>
```
vjkoskela: Can we use one of the JDK properties here?

rpm/scripts/post.sh
```diff
@@ -0,0 +1,5 @@
+/sbin/chkconfig --add mad
```
vjkoskela: These scripts don't need #!?

rpm/scripts/preun.sh
```diff
@@ -0,0 +1,5 @@
+if [ "$1" = 0 ]; then
+  /sbin/service mad stop > /dev/null 2>&1
+  /sbin/chkconfig --del mad
+fi
+exit 0
```
vjkoskela: Why the explicit exit? (versus the other scripts)

rpm/fs/etc/init.d/mad
BrandonArp: No, sh is assumed to be on all *nix installs.

rpm/scripts/post.sh
BrandonArp: Nope, they do not.

vjkoskela:
Looks very clean. A few things to consider as this proliferates:

* Move the profile into the parent pom.
* Move the plugin dependency into the plugin management block in the parent pom.

Otherwise, just a few questions. If they lead to change I can take another look, but I'm guessing its either just clarification and a ship-it.

pom.xml
BrandonArp: Done

rpm/scripts/preun.sh
BrandonArp: We don't want to fail pre-uninstall if we can't delete the service profile (and rollback the removal transaction).  Same with the post-uninstall.

rpm/fs/etc/init.d/mad
BrandonArp: Changed to sh.

PR #16, opened 2016-06-22
Initial Aggregator Changes to Support Arbitrary Key Sets

vjkoskela:
Extend keys from cluster-service-host to "all" dimensions and plumb (simplifying) changes through codebase.


src/main/java/com/arpnetworking/tsdcore/sinks/AggregationServerSink.java
```diff
@@ -50,25 +50,25 @@ public void recordAggregateData(final PeriodicData periodicData) {
...
-        final AggregatedData firstElement = dataList.iterator().next();
+    private Messages.StatisticSetRecord serializeMetricData(
+            final PeriodicData periodicData,
+            final String metricName,
+            final Collection<AggregatedData> data) {
 
+        // Map all dimensions
         final List<Messages.DimensionEntry> dimensions = Lists.newArrayList();
-        for (final Map.Entry<String, String> entry : periodicData.getDimensions().entrySet()) {
+        for (final Map.Entry<String, String> entry : periodicData.getDimensions().getDimensions().entrySet()) {
```
BrandonArp: This feels very weird.  Can you put a TODO in here to clean up the .getDimensions().getDimensions() call?

src/main/java/com/arpnetworking/tsdcore/statistics/TPStatistic.java
```diff
@@ -93,13 +92,7 @@ public Quantity calculate(final List<Quantity> orderedValues) {
...
     @Override
     public Quantity calculateAggregations(final List<AggregatedData> aggregations) {
-        final List<Quantity> allSamples = Lists.newArrayList();
-        for (final AggregatedData aggregation : aggregations) {
-            allSamples.addAll(aggregation.getSamples());
-        }
-        Collections.sort(allSamples);
-        final int index = (int) (Math.ceil((_percentile / 100.0) * (allSamples.size() - 1)));
-        return allSamples.get(index);
+        throw new UnsupportedOperationException("Unsupported operation: calculateAggregations(List<AggregatedData>)");
```
BrandonArp: Why not just remove all the calculateAggregations?  That is now the job of cagg in a completely separate code base.

BrandonArp:
Generally looks good. I'm excited for us to implement additional dimensions.  I'm already running into needing tags for data (not quite dimensions, but close).

src/main/java/com/arpnetworking/tsdcore/sinks/AggregationServerSink.java
vjkoskela: I just renamed the method in Set from _getDimensions_ to _getParameter_. Does that work for you?

src/main/java/com/arpnetworking/tsdcore/statistics/TPStatistic.java
vjkoskela: I didn't want to go down any unnecessary refactoring code paths in a single change. This already included clean-up for AggregatedData and PeriodicData which as you can tell was quite a sprawl. The statistics also need to be extracted into a shared library so I didn't want the copy in MAD to diverge too much from that in CAgg.

vjkoskela:
Thanks! As discussed earlier today it should not be necessary to inject additional dimensions in the colo case. The colo-specific CAgg instance can use the domain injecting sink to parse the domain from the hostname and add it as a dimension. Although you could do this in MAD relatively easily as well, please consider that we are trying to remove all embedded configuration from MAD.

PR #17, opened 2016-06-23
change cluster, service, host dimensions to drop _

BrandonArp:


matthayter:
whoa. Is this accompanied by a major version bump?

BrandonArp:
The "_" was only added to the dimension name with Ville's last change and was never externally visible.  Cagg accepts both the _ and non-_ versions.  This change just goes back to the original non-_ versions for the dimension keys.

PR #20, opened 2016-06-28
Fixes some flakey tests, add another test

BrandonArp:


src/main/java/com/arpnetworking/metrics/common/tailer/StatefulTailer.java
```diff
@@ -262,18 +263,7 @@ private void readLoop(final SeekableByteChannel reader) throws IOException, Inte
...
-                        rotate(
-                                Optional.of(reader),
-                                String.format(
-                                        "File rotation detected based last modified time access failure; file=%s",
-                                        _file));
-
-                        // Return to the file loop
-                        return;
-                    }
+                    lastChecked = Optional.of(attributes.getLastModifiedTime());
```
vjkoskela: How are you guaranteed that this won't fail if the file does not exist? Is the attributes one the new way of doing this?

src/test/java/com/arpnetworking/metrics/common/sources/FileSourceTest.java
```diff
@@ -868,7 +869,11 @@ private void truncate(final Path file) {
         try {
             FileChannel.open(file, StandardOpenOption.WRITE).truncate(0L).close();
         } catch (final IOException e) {
-            // Ignore
+            LOGGER.error()
```
vjkoskela: Should this fail the test?

BrandonArp: Agreed. Lets just let the exception escape and it'll properly fail the test.

src/main/java/com/arpnetworking/metrics/common/tailer/StatefulTailer.java
BrandonArp: Well, lets add a test for it.

BrandonArp: Actually fixed a test that I believe was intended to test for this case.  The line where attributes is set catches the same exception and does the same rotate.  Everything is good here.

BrandonArp:
Back to you.

vjkoskela:
Looks good to me. Waiting for build to complete.

BrandonArp:
Added another (random) test to bump branch coverage a little bit.

PR #21, opened 2016-07-14
Host, Service and Cluster to Annotations

vjkoskela:
Moved host, service and cluster from Record attributes into Record annotations. User annotations by the same names will be suppressed.

### DIFF
--- a/src/main/java/com/arpnetworking/metrics/mad/parsers/CollectdJsonToRecordParser.java
+++ b/src/main/java/com/arpnetworking/metrics/mad/parsers/CollectdJsonToRecordParser.java
@@ -35,6 +35,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.module.afterburner.AfterburnerModule;
 import com.google.common.base.Strings;
 import com.google.common.collect.HashMultimap;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Multimap;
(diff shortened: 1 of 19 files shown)


src/main/java/com/arpnetworking/metrics/mad/parsers/CollectdJsonToRecordParser.java
```diff
@@ -75,22 +76,17 @@
...
 
                 metricTags.put(Key.HOST_DIMENSION_KEY, record.getHost());
-                builder.setHost(record.getHost())
+                final DefaultRecord.Builder builder = new DefaultRecord.Builder()
                         .setId(UUID.randomUUID().toString())
                         .setTime(record.getTime())
-                        .setAnnotations(metricTags)
-                        .setCluster(cluster)
-                        .setService(service);
+                        .setAnnotations(ImmutableMap.copyOf(metricTags));
```
BrandonArp: I think this is going to double copy it.

src/main/java/com/arpnetworking/metrics/mad/parsers/CollectdJsonToRecordParser.java
```diff
@@ -111,7 +107,7 @@
                         .entrySet()
                         .stream()
                         .collect(Collectors.toMap(Map.Entry::getKey, CollectdJsonToRecordParser::mergeMetrics));
-                builder.setMetrics(collectedMetrics);
```
BrandonArp: I think this is also a double copy.

src/main/java/com/arpnetworking/metrics/mad/parsers/CollectdJsonToRecordParser.java
BrandonArp: Nvm, the docs say it will try to avoid copying.

src/main/java/com/arpnetworking/metrics/mad/parsers/CollectdJsonToRecordParser.java
BrandonArp: Disregard.

BrandonArp:
Thanks!

PR #28, opened 2016-09-02
fix dropped messages due to not accepting collectd data

BrandonArp:
Collectd sends us data like this.  We should accept it.


vjkoskela:
Nice.

PR #30, opened 2016-09-27
fix concurrency problem with telemetry actors/messages

BrandonArp:
Had this little gem blow up on me today.  The root of the problem is that the map is passed into a message and sent to the connection message processors.  This causes a concurrent modification exception when new metrics are being sent created while the metrics list is being walked to send to a websocket client.


vjkoskela:
Using locking across actors seems like the wrong way to fix this long term. Happy to patch it, but what's the plan here? Can you file an issue for it?

BrandonArp:
Given how often this code is called, I'm just going to copy it into an immutable map.  Will have updated PR shortly.

BrandonArp:
Updated. I think this is more in line with what we want.

PR #33, opened 2016-10-04
Add dimensions support

PhoenixRion:
Based on using 2g file format
Original PR https://github.com/ArpNetworking/metrics-aggregator-daemon/pull/27


src/main/java/com/arpnetworking/metrics/mad/parsers/JsonToRecordParser.java
```diff
@@ -586,17 +599,51 @@ private JsonToRecordParser(final Builder builder) {
...
     };
 
-    private static final String PREFIXED_HOST_KEY = "_host";
-    private static final String PREFIXED_SERVICE_KEY = "_service";
-    private static final String PREFIXED_CLUSTER_KEY = "_cluster";
+    private static final Function<Version2g.Sample, Quantity> VERSION_2G_SAMPLE_TO_QUANTITY = sample -> {
+        if (sample != null) {
+            if (Double.isFinite(sample.getValue())) {
+                final Unit sampleUnit = sample.getUnit2g() != null
+                        ? CompoundUnit.getExistingUnit(Iterables.getFirst(sample.getUnit2g().getNumerators(), null))
```
PhoenixRion: @vjkoskela "I'll have to take a closer look at this, but I want to stress that we have a good separation between the serialized types and the internal ones. This class is the mapping layer."

src/test/java/com/arpnetworking/metrics/mad/parsers/JsonToRecordParserV2gTest.java
```diff
@@ -0,0 +1,429 @@
...
+        }
+    }
+
+    private static Record parseRecord(final String fileName) throws ParsingException, IOException {
+        return new JsonToRecordParser.Builder()
+                .build()
+                .parse(Resources.toByteArray(Resources.getResource(
+                        JsonToRecordParserV2gTest.class, fileName)));
+    }
+}
```
PhoenixRion: @vjkoskela "Please revIew coverage for this code block as well."

PhoenixRion:
Copied over the two comments that I have not responded to in the original PR

vjkoskela:
Thanks for copying over the open comments from the internal PR. Please respond to the comments when they have been addressed. It also looks like you have some checkstyle violations.

https://s3.amazonaws.com/archive.travis-ci.org/jobs/160019477/log.txt

PhoenixRion:
@vjkoskela Any idea where this error is coming from " /home/travis/.m2/settings-security.xml (No such file or directory)" ?

mvn verify runs successfully on my local machine.

src/main/java/com/arpnetworking/metrics/mad/parsers/JsonToRecordParser.java
PhoenixRion: I moved the internal unit mapping out of this function.

src/test/java/com/arpnetworking/metrics/mad/parsers/JsonToRecordParserV2gTest.java
PhoenixRion: Added a test with empty dimensions as valid.

vjkoskela:
@PhoenixRion seems pretty clear to me it's coming from the Docker Hub interaction; namely, that the build can't pull down the base image.

```
[INFO] --- docker-maven-plugin:0.16.7:build (default-build) @ metrics-aggregator-daemon ---
[ERROR] DOCKER> Cannot decrypt password: org.sonatype.plexus.components.sec.dispatcher.SecDispatcherException: org.sonatype.plexus.components.sec.dispatcher.SecDispatcherException: java.io.FileNotFoundException: /home/travis/.m2/settings-security.xml (No such file or directory)
```

The latest build on master succeeds and so does Brandon's PR #37. Have you checked that your changes are properly rebased?

@BrandonArp the only thing I can think of is that there is a problem decrypting the secret because this is a forked PR. Do you think that's possible?

src/main/java/com/arpnetworking/tsdcore/model/CompositeUnit.java
```diff
@@ -0,0 +1,338 @@
...
+ */
+package com.arpnetworking.tsdcore.model;
+
+import com.google.common.base.Optional;
+import com.google.common.collect.ImmutableMap;
+import net.sf.oval.constraint.NotNull;
+
+
+/**
+ * Compound Unit used in 2g file format.
```
vjkoskela: I still don't understand why this is a part of the model. The serialization format and its model representation should be entirely encapsulated within that package and those classes. Brandon just did the same transformation from Protobuf in #37 without adding any of this  complexity. If I am missing some reason why the JSON representation is different, or how this advances the internal representation in some way please explain.

vjkoskela:
Brandon's change has been merged. Since your PR is adding dimensions to the model generated by Parsers you will need to update the HTTP parser as well to map dimensions from Protobuf into dimensions on the model. The other option would have been for Brandon to also modify the model which would have cause several merge conflicts, I think the changes his code will require are very minor.

src/main/java/com/arpnetworking/tsdcore/model/CompositeUnit.java
PhoenixRion: If the serialized format supports all of these possible units it seems good lay some ground work for being able to use more than the current set of units.  I don't have the time right now to fully implement all of the functionality, such as transforming between units or fully replacing the existing Unit class but this seemed like a good place to start.

PhoenixRion:
@vjkoskela @BrandonArp I don't know what to do about the settings-security.xml file.  Seems pretty likely that there is something related this PR being from a fork because the error was not present before.

src/main/java/com/arpnetworking/tsdcore/model/CompositeUnit.java
vjkoskela: Then I think you should explain your design, I am not clear on how this advances towards a unified representation because you are creating another representation inside the mad model. As I have already said there should only be two: the serialized form and the mad internal.

The current mad internal form is here:
https://github.com/ArpNetworking/metrics-aggregator-daemon/blob/master/src/main/java/com/arpnetworking/tsdcore/model/Unit.java

I don't have time to design the extension at this point either and without a clear plan explaining these changes I again suggest that you look at Brandon's recent change to support the same feature set over protobuf.

vjkoskela:
I don't have time to look at the settings issue right now. I would recommend cutting a noop PR with maven in debut mode.

vjkoskela:
Forked builds fixed with #39

src/main/java/com/arpnetworking/tsdcore/model/CompositeUnit.java
PhoenixRion: My idea was that the CompositeUnit class would be essentially UnitV2.  Breaking up the enumerations into this structure was intended to facilitate the transformation between any two units of the same type.

There is a decent chunk of work left to replace the existing Unit class which I have not had time to work on but it should be able to be a direct replacement.

vjkoskela: I think that's exactly where the disconnect is. The broken format (unit + scale) is a serialization format. There is no need, at least at the moment, to undertake a large refactor of the internal Unit class. In fact, this is precisely why we separate the internal and external data models; so changes to one do not ripple to the other.

Instead I strongly recommend that we simply map the 2G Unit into the existing internal data model. It would be nice if you added the missing units to the internal model since the 2G format is more expressive. This would be a simple task, in fact a Python or Bash code generator should be able to whip up the change in no time (otherwise you'll need to do some cut-and-paste or get Gil or Matt's help with editor macros).

The other points is the naming inside the model class, please don't call something "Unit2G" since it's already within the 2G Version class. These are all encapsulated and further disambiguation via naming is unnecessary.

matthayter:
What's the status here @PhoenixRion @vjkoskela ? Who is waiting on whom?

vjkoskela:
I am waiting for resolution on the unit issue, there should be a thread in that buried in here somewhere. The current proposal is to start a rewrite of units in MAD, but as Ryan has said he will not be able to complete it so I am not inclined to start it. Let me know if there is something else you need from me on this.

PhoenixRion:
I should have some time next week to work on updating this code.

vjkoskela:
No problem with me, as discussed 2g file support is low on our priority list because http already supports this.

PhoenixRion:
This is a diff with a full update for the existing Unit class.  https://github.com/PhoenixRion/metrics-aggregator-daemon/compare/master...PhoenixRion:update-unit-to-composit-structure

I thought it would be a good idea to share for comments before I go through the remaining effort of finishing up checkstyle and updating all the javadocs.  I have run the verification to execute all of the existing tests successfully.

src/main/java/com/arpnetworking/metrics/mad/model/json/Version2g.java
```diff
@@ -0,0 +1,394 @@
...
+            private Unit(final Builder builder) {
+                _numerators = builder._numerators == null
+                        ? ImmutableList.of()
+                        : ImmutableList.copyOf(builder._numerators);
+                _denominators = builder._denominators == null
+                        ? ImmutableList.of()
+                        : ImmutableList.copyOf(builder._denominators);
+            }
+
+            private final List<CompositeUnit> _numerators;
```
vjkoskela: These should be stored and returned as ImmutableList so callers don't need to defensively copy.

src/main/java/com/arpnetworking/metrics/mad/model/json/Version2g.java
```diff
@@ -0,0 +1,394 @@
...
+                 *
+                 * @param value Value
+                 * @return This builder
+                 */
+                public Builder setDenominators(final List<CompositeUnit> value) {
+                    _denominators = value;
+                    return this;
+                }
+
+                private List<CompositeUnit> _numerators;
```
vjkoskela: The file format model can NOT reference the internal model. Please reach out offline if this is not clear (it has been mentioned on every iteration of this change).

src/main/java/com/arpnetworking/metrics/mad/model/json/Version2g.java
```diff
@@ -0,0 +1,394 @@
...
+                private List<CompositeUnit> _denominators;
+            }
+        }
+    }
+
+    /**
+     * Represents a counter, timer, or gauge element.  Includes a list of
+     * samples.
+     */
+    public static final class Element {
```
vjkoskela: Why is unit inside sample and sample not inside element? My preference is to not nest these classes within each other.

src/main/java/com/arpnetworking/metrics/mad/parsers/JsonToRecordParser.java
```diff
@@ -430,6 +420,29 @@ private static void putVariablesVersion2fSteno(
...
     }
 
+    private static void putVariablesVersion2g(
+            final Map<String, Version2g.Element> elements,
+            final MetricType metricKind,
+            final ImmutableMap.Builder<String, Metric> variables) {
+
+        for (final Map.Entry<String, Version2g.Element> entry : elements.entrySet()) {
+            final Version2g.Element element = entry.getValue();
+            final List<Quantity> quantities = Lists.newArrayList(
```
vjkoskela: There seems to be some unnecessary copying going on here.

src/main/java/com/arpnetworking/tsdcore/model/CompositeUnit.java
```diff
@@ -0,0 +1,338 @@
...
+ */
+package com.arpnetworking.tsdcore.model;
+
+import com.google.common.base.Optional;
+import com.google.common.collect.ImmutableMap;
+import net.sf.oval.constraint.NotNull;
+
+
+/**
+ * Composite Unit used in 2g file format.
```
vjkoskela: No. You can NOT bleed the file model into the internal model. The entire point of the design is to keep them separate so changes do NOT ripple across the codebase.

src/main/java/com/arpnetworking/tsdcore/model/CompositeUnit.java
```diff
@@ -0,0 +1,338 @@
...
+    private Scale _scale;
+
+    /**
+     * Get the existing unit that maps to the covmmpound unit.
+     *
+     * @param compositeUnit
+     * ompoundUnit newer unit format.
+     * @return existing unit which maps to the newer format
+     */
+    public static Unit getLegacyUnit(final Optional<CompositeUnit> compositeUnit) {
```
vjkoskela: This cannot return null.

src/main/java/com/arpnetworking/tsdcore/model/CompositeUnit.java
```diff
@@ -0,0 +1,338 @@
...
+    enum Category {
+        TIME,
+        DATA_SIZE,
+        TEMPERATURE,
+        ROTATION
+    }
+
+    @NotNull
+    private Type _type;
+    private Scale _scale;
```
vjkoskela: If this is nullable it should be optional.

src/main/java/com/arpnetworking/tsdcore/model/CompositeUnit.java
```diff
@@ -0,0 +1,338 @@
...
+         */
+        CENTI(10 ^ -2),
+        /**
+         * 10^-1.
+         */
+        DECI(10 ^ -1),
+        /**
+         * 1.
+         */
+        ONE(1),
```
vjkoskela: Since it's unclear whether this is an internal model or a format model, my comment may or
May not be correct: but ONE afaik is not a valid value in the scheme. Right?

src/main/java/com/arpnetworking/tsdcore/model/CompositeUnit.java
```diff
@@ -0,0 +1,338 @@
...
+     * @return existing unit which maps to the newer format
+     */
+    public static Unit getLegacyUnit(final Optional<CompositeUnit> compositeUnit) {
+        return compositeUnit.isPresent()
+                ? LEGACY_UNIT_MAP.getOrDefault(compositeUnit.get(), null)
+                : null;
+    }
+
+    private static final ImmutableMap<CompositeUnit, Unit> LEGACY_UNIT_MAP = new ImmutableMap.Builder<CompositeUnit, Unit>()
+            .put(new CompositeUnit(Scale.NANO, Type.SECOND), Unit.NANOSECOND)
```
vjkoskela: As I mentioned in the previous review, Brandon did this by simply concatenation the scale and unit. While not pretty as far as I know it worked just fine. I'm happy to wrap that in a function to allow us to change the logic in one place. However this mapping should be in the parser classes.

src/main/java/com/arpnetworking/tsdcore/model/CompositeUnit.java
```diff
@@ -0,0 +1,338 @@
...
+    }
+
+    public void setScale(final Scale value) {
+        _scale = value;
+    }
+
+    /**
+     * Composite unit data type.
+     */
+    public enum Type {
```
vjkoskela: Where is this type used? There is already a type associated with each unit on the existing internal model and the file model does not as far as I can tell need this at all.

src/main/java/com/arpnetworking/tsdcore/model/Quantity.java
```diff
@@ -344,6 +344,17 @@ public Builder setUnit(final Unit value) {
         }
 
         /**
+         * Set the unit. Optional. Default is no unit.
+         *
+         * @param value The composite unit.
+         * @return This <code>Builder</code> instance.
+         */
+        public Builder setUnitByComposite(final CompositeUnit value) {
```
vjkoskela: No.

matthayter:
Hey all, I'll be working to get this merged/resolved. I believe @BrandonArp has taken over handling the feeback for this PR - if that's the case would you @BrandonArp be available for a voice chat tomorrow afternoon, the w/end or early next week?

src/main/java/com/arpnetworking/metrics/mad/model/DefaultRecord.java
```diff
@@ -181,19 +189,16 @@ public Builder setAnnotations(final ImmutableMap<String, String> value) {
             return this;
         }
 
-        // Called by OVal reflectively
-        @SuppressFBWarnings("UPM_UNCALLED_PRIVATE_METHOD")
-        private boolean validateAnnotations(final ImmutableMap<String, String> annotations) {
```
matthayter: Is the intent to no longer verify that the host, service and cluster exist in the Record?

src/main/java/com/arpnetworking/metrics/mad/parsers/CollectdJsonToRecordParser.java
```diff
@@ -87,7 +87,8 @@
                 final DefaultRecord.Builder builder = new DefaultRecord.Builder()
                         .setId(UUID.randomUUID().toString())
                         .setTime(record.getTime())
-                        .setAnnotations(ImmutableMap.copyOf(metricTags));
+                        .setAnnotations(ImmutableMap.copyOf(metricTags))
+                        .setDimensions(ImmutableMap.copyOf(metricTags));
```
matthayter: Gotta verify whether setting both to `metricTags` is correct

src/main/java/com/arpnetworking/metrics/mad/parsers/JsonToRecordParser.java
```diff
@@ -137,27 +142,18 @@ public Record parse(final byte[] data) throws ParsingException {
...
+        dimensionsBuilder.put(Key.HOST_DIMENSION_KEY, _defaultHost);
+        dimensionsBuilder.put(Key.SERVICE_DIMENSION_KEY, _defaultService);
+        dimensionsBuilder.put(Key.CLUSTER_DIMENSION_KEY, _defaultCluster);
 
         return new DefaultRecord.Builder()
                 .setMetrics(variables.build())
                 .setId(UUID.randomUUID().toString())
                 .setTime(timestamp)
-                .setAnnotations(mappedAnnotations.build())
+                .setAnnotations(ImmutableMap.copyOf(annotations.getOtherAnnotations()))
```
matthayter: This will now include the annotations excluded on previous lines 149-151 - is this intended?

src/main/java/com/arpnetworking/metrics/mad/parsers/JsonToRecordParser.java
BrandonArp: Agreed, get rid of the Lists.newArrayList wrapping call.  The streams collector will create that list for you.

src/main/java/com/arpnetworking/metrics/mad/parsers/JsonToRecordParser.java
```diff
@@ -586,17 +649,51 @@ private JsonToRecordParser(final Builder builder) {
         }
     };
 
-    private static final String PREFIXED_HOST_KEY = "_host";
-    private static final String PREFIXED_SERVICE_KEY = "_service";
-    private static final String PREFIXED_CLUSTER_KEY = "_cluster";
+    private static final Function<Version2g.Sample, Quantity> VERSION_2G_SAMPLE_TO_QUANTITY = sample -> {
```
BrandonArp: Please just make this a function instead of a lambda.

src/main/java/com/arpnetworking/metrics/mad/parsers/ProtobufToRecordParser.java
```diff
@@ -125,9 +126,14 @@ private Unit baseUnit(final ClientV1.CompoundUnit compoundUnit) {
         for (final ClientV1.AnnotationEntry annotationEntry : record.getAnnotationsList()) {
             annotations.put(annotationEntry.getName(), annotationEntry.getValue());
         }
+        return annotations.build();
+    }
+
+    private ImmutableMap<String, String> buildDimensinos(final ClientV1.Record record) {
```
BrandonArp: Typo in name.

src/main/java/com/arpnetworking/metrics/mad/parsers/JsonToRecordParser.java
matthayter: Happy to - shall I make the other formats' Lambdas into functions also?
```
VERSION_2C_SAMPLE_TO_QUANTITY
VERSION_2D_SAMPLE_TO_QUANTITY
VERSION_2E_SAMPLE_TO_QUANTITY
VERSION_2F_SAMPLE_TO_QUANTITY
VERSION_2F_STENO_SAMPLE_TO_QUANTITY
```

matthayter:
@BrandonArp still waiting on this one when/if you have some time

BrandonArp:
Sorry, I thought I was still waiting on you for something.  Thanks for the gentle nudge.

PR #37, opened 2016-10-10
Support application HTTP sink

BrandonArp:


src/main/java/com/arpnetworking/http/Routes.java
```diff
@@ -156,22 +156,29 @@ public Routes(
...
-                    .resolveOne(FiniteDuration.create(1, TimeUnit.SECONDS));
-            return FutureConverters.toJava(refFuture).thenCompose(
-                    ref -> {
-                        final CompletableFuture<HttpResponse> response = new CompletableFuture<>();
-                        ref.tell(new RequestReply(request, response), ActorRef.noSender());
-                        return response;
-                    })
-                    .exceptionally(err -> HttpResponse.create().withStatus(404));
+        } else if (HttpMethods.POST.equals(request.method())) {
+            if (path.equals(COLLECTD_V1_SOURCE_PREFIX) || path.equals("/source/collectd")) {
```
vjkoskela: "/source/collectd" => psf

src/main/java/com/arpnetworking/http/Routes.java
```diff
@@ -156,22 +156,29 @@ public Routes(
...
+    private CompletionStage<HttpResponse> dispatchHttpRequest(final HttpRequest request, final String actorName) {
+        final Future<ActorRef> refFuture = _actorSystem.actorSelection(actorName)
+                .resolveOne(FiniteDuration.create(1, TimeUnit.SECONDS));
+        return FutureConverters.toJava(refFuture).thenCompose(
+                ref -> {
+                    final CompletableFuture<HttpResponse> response = new CompletableFuture<>();
+                    ref.tell(new RequestReply(request, response), ActorRef.noSender());
+                    return response;
+                })
+                .exceptionally(err -> HttpResponse.create().withStatus(404));
```
vjkoskela: 404? -> we need to be able to distinguish between a 404 (bad payload) and an internal server error.

src/main/java/com/arpnetworking/metrics/common/sources/ClientHttpSourceV1.java
```diff
@@ -0,0 +1,57 @@
...
+ * limitations under the License.
+ */
+package com.arpnetworking.metrics.common.sources;
+
+import com.arpnetworking.metrics.mad.parsers.IsmClientProtocolRecordParser;
+
+/**
+ * Processes HTTP requests from the metrics client, extracts data and emits metrics.
+ *
+ * @author Brandon Arp (brandon dot arp at inscopemetrics dot com)
```
vjkoskela: Just want to confirm that the copyright and contact info reflect the context under which this was completed.

src/main/java/com/arpnetworking/metrics/common/sources/ClientHttpSourceV1.java
```diff
@@ -0,0 +1,57 @@
...
+     * Protected constructor.
+     *
+     * @param builder Instance of <code>Builder</code>.
+     */
+    private ClientHttpSourceV1(final Builder builder) {
+        super(builder);
+    }
+
+    /**
+     * CollectdHttpSourceV1 {@link BaseSource.Builder} implementation.
```
vjkoskela: Copy-paste error.

src/main/java/com/arpnetworking/metrics/common/sources/ClientHttpSourceV1.java
```diff
@@ -0,0 +1,57 @@
...
+     * @author Brandon Arp (brandon dot arp at inscopemetrics dot com)
+     */
+    public static final class Builder extends HttpSource.Builder<Builder, ClientHttpSourceV1> {
+        /**
+         * Public constructor.
+         */
+        public Builder() {
+            super(ClientHttpSourceV1::new);
+            setActorName("appv1");
+            setParser(new IsmClientProtocolRecordParser());
```
vjkoskela: The other parsers are named "CollectdJsonToRecordParser" and "JsonToRecordParser". Might I suggest one of:

* IsmProtobufToRecordParser
* ProtobufToRecordParser

The last one is a parallel to JsonToRecordParser and although currently under separate domains the plan is still to rehome most of this code there and have those protocols become the public standard.

src/main/java/com/arpnetworking/metrics/common/sources/ClientHttpSourceV1.java
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        @Override
```
vjkoskela: Missing Javadoc.

src/main/java/com/arpnetworking/metrics/common/sources/CollectdHttpSourceV1.java
```diff
@@ -45,6 +45,7 @@ private CollectdHttpSourceV1(final Builder builder) {
         public Builder() {
             super(CollectdHttpSourceV1::new);
             setParser(new CollectdJsonToRecordParser());
+            setActorName("collectdv1");
```
vjkoskela: Should this be a public static final somewhere since it's referred to at least in two places?

src/main/java/com/arpnetworking/metrics/mad/Main.java
```diff
@@ -250,14 +250,15 @@ private Injector launchGuice(final ActorSystem actorSystem) {
         }
 
         // Instantiate the metrics factory
+        final String sinkHost = "0.0.0.0".equals(_configuration.getHttpHost()) ? "localhost" : _configuration.getHttpHost();
```
vjkoskela: Should we map 0.0.0.0 to 127.0.0.1 instead of the name localhost? Not sure it matters.

src/main/java/com/arpnetworking/metrics/mad/parsers/IsmClientProtocolRecordParser.java
```diff
@@ -0,0 +1,129 @@
...
+import java.util.UUID;
+import javax.annotation.Nullable;
+
+/**
+ * Parses the Inscope Metrics protobuf binary protocol into records.
+ *
+ * @author Brandon Arp (brandon dot arp at inscopemetrics dot com)
+ */
+public class IsmClientProtocolRecordParser implements Parser<List<Record>, HttpRequest> {
+    @Override
```
vjkoskela: Missing javadoc.

src/main/java/com/arpnetworking/metrics/mad/parsers/IsmClientProtocolRecordParser.java
```diff
@@ -0,0 +1,129 @@
...
+    public List<Record> parse(final HttpRequest data) throws ParsingException {
+        try {
+            final ClientV1.RecordSet request = ClientV1.RecordSet.parseFrom(data.getBody());
+            final List<Record> records = Lists.newArrayList();
+            for (ClientV1.Record record : request.getRecordsList()) {
```
vjkoskela: record should be final

src/main/java/com/arpnetworking/metrics/mad/parsers/IsmClientProtocolRecordParser.java
```diff
@@ -0,0 +1,129 @@
...
+                        .setId(new UUID(high, low).toString())
+                        .setTime(new DateTime(record.getEndMillisSinceEpoch()))
+                        .setAnnotations(buildAnnotations(record))
+                        .setMetrics(buildMetrics(record));
+
+                records.add(builder.build());
+            }
+            return records;
+        } catch (final InvalidProtocolBufferException e) {
+            throw new ParsingException("Could not create Request message from data", data.getBody(), e);
```
vjkoskela: So we should map only ParsingException to 404 in the handler.

vjkoskela: ^ Can you double check that the CollectD Http endpoint behaves similarly?

src/main/java/com/arpnetworking/metrics/mad/parsers/IsmClientProtocolRecordParser.java
```diff
@@ -0,0 +1,129 @@
...
+        processEntries(metrics, record.getTimersList(), MetricType.TIMER);
+        processEntries(metrics, record.getGaugesList(), MetricType.GAUGE);
+        return metrics.build();
+    }
+
+    private void processEntries(
+            final ImmutableMap.Builder<String, Metric> metrics,
+            final List<ClientV1.MetricEntry> entries,
+            final MetricType metricType) {
+        for (ClientV1.MetricEntry metricEntry : entries) {
```
vjkoskela: final

src/main/java/com/arpnetworking/metrics/mad/parsers/IsmClientProtocolRecordParser.java
```diff
@@ -0,0 +1,129 @@
...
+            final DefaultMetric.Builder metricBuilder = new DefaultMetric.Builder()
+                    .setType(metricType);
+            final List<Quantity> quantities = Lists.newArrayList();
```
vjkoskela: Can you right-size the array list based on  metricEntry.getSamplesList()?

src/main/java/com/arpnetworking/metrics/mad/parsers/IsmClientProtocolRecordParser.java
```diff
@@ -0,0 +1,129 @@
...
+            for (ClientV1.DoubleQuantity quantity : metricEntry.getSamplesList()) {
```
vjkoskela: final

src/main/java/com/arpnetworking/metrics/mad/parsers/IsmClientProtocolRecordParser.java
```diff
@@ -0,0 +1,129 @@
...
+            }
+
+            metricBuilder.setValues(quantities);
+            metrics.put(metricEntry.getName(), metricBuilder.build());
+        }
+    }
+
+    @Nullable
+    private Unit baseUnit(final ClientV1.CompoundUnit compoundUnit) {
+        if (compoundUnit.getNumeratorCount() > 0) {
```
vjkoskela: ! isEmpty?

src/main/java/com/arpnetworking/metrics/mad/parsers/IsmClientProtocolRecordParser.java
```diff
@@ -0,0 +1,129 @@
...
+            final ClientV1.Unit selectedUnit = compoundUnit.getNumerator(0);
+            if (selectedUnit.getType() == ClientV1.Unit.Type.UNRECOGNIZED) {
```
vjkoskela: I would prefer .equsls, with the constant on the left side. == always assumes something about the implementation (e.g. enum vs static instance).

src/main/java/com/arpnetworking/metrics/mad/parsers/IsmClientProtocolRecordParser.java
```diff
@@ -0,0 +1,129 @@
...
+                return null;
+            }
+            final String unitName;
+            if (selectedUnit.getScale() != ClientV1.Unit.Scale.UNIT && selectedUnit.getScale() != ClientV1.Unit.Scale.UNRECOGNIZED) {
```
vjkoskela: Same here.

src/main/java/com/arpnetworking/metrics/mad/parsers/IsmClientProtocolRecordParser.java
```diff
@@ -0,0 +1,129 @@
...
+        return null;
+    }
+
+    private ImmutableMap<String, String> buildAnnotations(final ClientV1.Record record) {
+        final ImmutableMap.Builder<String, String> annotations = ImmutableMap.builder();
+        for (ClientV1.AnnotationEntry annotationEntry : record.getAnnotationsList()) {
+            annotations.put(annotationEntry.getName(), annotationEntry.getValue());
+        }
+        for (ClientV1.DimensionEntry dimensionEntry : record.getDimensionsList()) {
+            annotations.put(dimensionEntry.getName(), dimensionEntry.getValue());
```
vjkoskela: Why are we mapping dimensions into annotations? Doesn't the internal model support dimensions separately?

vjkoskela:
Very nice and simple.

src/main/java/com/arpnetworking/metrics/common/sources/ClientHttpSourceV1.java
BrandonArp: Confirmed.  This is ISM.

src/main/java/com/arpnetworking/http/Routes.java
BrandonArp: Sure

src/main/java/com/arpnetworking/metrics/mad/parsers/IsmClientProtocolRecordParser.java
BrandonArp: Invalid data maps to 400's.  There are tests for this.

src/main/java/com/arpnetworking/metrics/common/sources/ClientHttpSourceV1.java
BrandonArp: Done

src/main/java/com/arpnetworking/metrics/common/sources/ClientHttpSourceV1.java
BrandonArp: Fixed.

src/main/java/com/arpnetworking/metrics/common/sources/ClientHttpSourceV1.java
BrandonArp: Fixed.

src/main/java/com/arpnetworking/metrics/mad/Main.java
BrandonArp: Nope.  Localhost was quite specific.  127.0.0.0/8 is the local range.  I've seen 127.0.1.1 be bound as well.  I think it's friendlier to use "localhost" and let the owner of the box delegate which IP address it uses.

src/main/java/com/arpnetworking/metrics/mad/parsers/IsmClientProtocolRecordParser.java
BrandonArp: One of these days I'm gonna delete all of those.  Might be my very next PR....

src/main/java/com/arpnetworking/metrics/mad/parsers/IsmClientProtocolRecordParser.java
BrandonArp: Fixed.

src/main/java/com/arpnetworking/metrics/mad/parsers/IsmClientProtocolRecordParser.java
BrandonArp: Fixed

src/main/java/com/arpnetworking/metrics/mad/parsers/IsmClientProtocolRecordParser.java
BrandonArp: Sure.

src/main/java/com/arpnetworking/metrics/mad/parsers/IsmClientProtocolRecordParser.java
BrandonArp: Nope.  Record does not support dimensions yet.

src/main/java/com/arpnetworking/metrics/common/sources/CollectdHttpSourceV1.java
BrandonArp: Sure.

src/main/java/com/arpnetworking/http/Routes.java
vjkoskela: This did not seem to get updated.

src/main/java/com/arpnetworking/http/Routes.java
vjkoskela: Got it. I'm not sure that the actor resolution failure should be a 404. It has nothing to do with the inputs, right? So it's beyond the caller's ability to control.

src/main/java/com/arpnetworking/http/Routes.java
vjkoskela: Looks like you made the psf :)

```ACTOR_APP_V1```

src/main/java/com/arpnetworking/metrics/mad/Main.java
vjkoskela: Fair enough. The other alternative is to allow it to be configurable separately. I think this is fine for now.

src/main/java/com/arpnetworking/http/Routes.java
BrandonArp: This is partially true.  The actor is only started if there is a pipeline that starts it up.  So, it may be there or it may not.  We should talk about how pipelines work in the new world.

src/main/java/com/arpnetworking/metrics/mad/parsers/IsmClientProtocolRecordParser.java
vjkoskela: Right, so this is more work for Ryan.

src/main/java/com/arpnetworking/http/Routes.java
vjkoskela: Sure, but even if the pipeline does not start-up it's not the callers fault that happened. The cause may very well be MAD misconfiguration (or whatever) but we can't assume the caller is responsible for that. What am I missing here that the request is somehow tied based on its parameters to the actor?

BrandonArp: 404 is not found.  In this case, the route doesn't exist because that actor hasn't been launched to serve it.

PR #40, opened 2016-10-21
Thread pool to read file lines.

matthayter:
Strategy:
- Use a pool of n threads running n runnables in a loop reading from the blocking work queue
- Runnables try to never die, and are not restarted if they do
- Runnables catch `RuntimeException`, logging and continuing
- Runnables catch `Throwable` afterwards, and log and exit - this is consided the extreme case where it is inappropriate to continue
- Runnables may exit by flipping a shared atomic boolean
- The thread reading the file blocks if the queue is full and waits for space.

Not sure how you guys feel; this pattern is of course ripe for abstraction (e.g. `WorkQueueExecutor<WorkItem>(Consumer c, int nWorkers).accept(WorkItem w)` ) but it may be a bit premature for that, and at that point perhaps we may as well do actors.

Note: the order in which `Records` are bucketed is now non-deterministic. Anyone know of any assumptions of ordered-datetime in the query log?

### DIFF
--- a/src/main/java/com/arpnetworking/metrics/common/sources/FileSource.java
+++ b/src/main/java/com/arpnetworking/metrics/common/sources/FileSource.java
@@ -206,7 +206,7 @@ public final class FileSource<T> extends BaseSource {
         public void handle(final byte[] line) {
             try {
                 _lineQueue.put(line);
-            } catch (InterruptedException e) {
+            } catch (final InterruptedException e) {
                 _logger.error()
                         .setMessage("FileSource tailer thread interrupted while waiting to write line to queue")
                         .addData("source", FileSource.this)
@@ -257,9 +257,13 @@ public final class FileSource<T> extends BaseSource {
 
     }
 
-    private static class LineQueueRunner<T> implements Runnable {
+    private static final class LineQueueRunner<T> implements Runnable {
 
-        private LineQueueRunner(final AtomicBoolean lineProcessingStopped, final Parser<T, byte[]> parser, final FileSource<T> fileSource, final ArrayBlockingQueue<byte[]> queue) {
+        private LineQueueRunner(
+                final AtomicBoolean lineProcessingStopped,
+                final Parser<T, byte[]> parser,
+                final FileSource<T> fileSource,
+                final ArrayBlockingQueue<byte[]> queue) {
             _lineProcessingStopped = lineProcessingStopped;
             _parser = parser;
             _fileSource = fileSource;
@@ -268,11 +272,11 @@ public final class FileSource<T> extends BaseSource {
 
         @Override
         public void run() {
-            while(!_lineProcessingStopped.get()) {
+            while (!_lineProcessingStopped.get()) {
                 final byte[] line;
                 try {
                     line = _queue.take();
-                } catch (InterruptedException e) {
+                } catch (final InterruptedException e) {
                     LOGGER.info()
                             .setMessage("Thread interrupted while waiting for query log line.")
                             .log();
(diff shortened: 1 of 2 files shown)


matthayter:
Dang, looks like we've got a flakey test; passes fine on my machine. I'll try a rebuild to attempt to verify.

src/main/java/com/arpnetworking/metrics/common/sources/FileSource.java
```diff
@@ -65,6 +70,8 @@ public void start() {
     public void stop() {
         _tailer.stop();
         _tailerExecutor.shutdown();
+        _lineProcessingStopped.set(true);
+        _lineProcessorExecutor.shutdownNow();
```
vjkoskela: Why shutdownNow versus shutdown and await?

src/main/java/com/arpnetworking/metrics/common/sources/FileSource.java
```diff
@@ -114,19 +129,34 @@ private FileSource(final Builder<T> builder) {
             positionStore = new FilePositionStore.Builder().setFile(builder._stateFile).build();
         }
 
+        _lineQueue = new ArrayBlockingQueue<>(builder._lineQueueMaxSize);
+
+        final LogTailerListener logTailerListener = new LogTailerListener();
```
vjkoskela: Why was this pulled out?

src/main/java/com/arpnetworking/metrics/common/sources/FileSource.java
```diff
@@ -73,6 +80,14 @@ public void stop() {
                     .setThrowable(e)
                     .log();
         }
+        try {
+            _lineProcessorExecutor.awaitTermination(10, TimeUnit.SECONDS);
```
vjkoskela: Oh you do await. My concern here is to ensure all data read from the file is processed so that we can do proper check pointing on shutdown. If we drop work at various stages it's going to make this MUCH harder.

src/main/java/com/arpnetworking/metrics/common/sources/FileSource.java
```diff
@@ -174,17 +204,16 @@ public void fileOpened() {
...
+                _lineQueue.put(line);
+            } catch (final InterruptedException e) {
                 _logger.error()
-                        .setMessage("Failed to parse data")
+                        .setMessage("FileSource tailer thread interrupted while waiting to write line to queue")
+                        .addData("source", FileSource.this)
                         .setThrowable(e)
                         .log();
-                return;
+                e.printStackTrace();
```
vjkoskela: Ahh, no.

src/main/java/com/arpnetworking/metrics/common/sources/FileSource.java
```diff
@@ -301,6 +395,31 @@ public Builder() {
         }
 
         /**
+         * Sets the number of threads to be used to process lines read from the file. Optional.
+         * Default is Runtime.getRuntime().availableProcessors().
+         *
+         * @param value The number of threads to be used.
+         * @return This instance of <code>Builder</code>.
+         */
+        public final Builder<T> setLineProcessorThreads(final int value) {
```
vjkoskela: One of the requirements we discussed was supporting "2C" which is two times cores, or more generally "{N}C". Please revise.

src/main/java/com/arpnetworking/metrics/common/sources/FileSource.java
```diff
@@ -301,6 +395,31 @@ public Builder() {
...
+            _lineProcessorThreads = value;
+            return this;
+        }
+
+        /**
+         * Sets the maximum size of the queue of lines to be stored in memory before processing. Optional.
+         * Reading of lines from file will block if the queue is full.
+         * Default is Runtime.getRuntime().availableProcessors() * 2.
```
vjkoskela: Why is the queue size based on he number of processors?

src/test/java/com/arpnetworking/metrics/common/sources/FileSourceTest.java
```diff
@@ -782,7 +787,8 @@ public void testTailerLogRotationCopyTruncateSmallToLarge() throws IOException,
                         .setSourceFile(file)
                         .setStateFile(state)
                         .setParser(_parser)
-                        .setInterval(Duration.millis(INTERVAL)),
+                        .setInterval(Duration.millis(INTERVAL))
+                        .setLineProcessorThreads(1),
```
vjkoskela: You've included no tests that exercise the concurrent functionality.

BrandonArp:
Closing for now as we do not recommend using file tailing as the way to ingest metrics.

PR #41, opened 2016-11-16
fix dropped metrics with multiple parts

BrandonArp:
also adds a metric for body size

### DIFF
--- a/src/main/java/com/arpnetworking/http/Routes.java
+++ b/src/main/java/com/arpnetworking/http/Routes.java
@@ -38,6 +38,7 @@ import akka.stream.javadsl.Sink;
 import akka.stream.javadsl.Source;
 import akka.util.ByteString;
 import akka.util.Timeout;
+import com.arpnetworking.metrics.Counter;
 import com.arpnetworking.metrics.Metrics;
 import com.arpnetworking.metrics.MetricsFactory;
 import com.arpnetworking.metrics.Timer;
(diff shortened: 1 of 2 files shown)


src/main/java/com/arpnetworking/http/Routes.java
```diff
@@ -240,6 +243,18 @@ private String createTimerName(final HttpRequest request) {
         return nameBuilder.toString();
     }
 
+    private String createRequestSizeCounterName(final HttpRequest request) {
+        final StringBuilder nameBuilder = new StringBuilder()
+                .append("rest_service/")
```
vjkoskela: Pull this out into a PSF; also used above and should be the same.

src/main/java/com/arpnetworking/http/Routes.java
```diff
@@ -240,6 +243,18 @@ private String createTimerName(final HttpRequest request) {
...
+                .append(request.method().value());
+        if (!request.getUri().path().startsWith("/")) {
+            nameBuilder.append("/");
+        }
+        nameBuilder.append(request.getUri().path());
```
vjkoskela: Seems like this is identical to the timer one above; refactor? Also, I don't think having a path be a metric and folder in ReMet is going to look the best. I think we should be explicit wrt latency.

BrandonArp:
Addressed, back to you.

PR #43, opened 2016-11-25
UnitMappingSink

vjkoskela:
Added unit mapping sink adapted from Cluster Aggregator for MAD data model. This should be a transient state until we normalize on ingest.

### DIFF
--- a/src/test/java/com/arpnetworking/tsdcore/sinks/UnitMappingSinkTest.java
+++ b/src/test/java/com/arpnetworking/tsdcore/sinks/UnitMappingSinkTest.java
@@ -1,0 +1,111 @@
+/**
+ * Copyright 2016 Inscope Metrics Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


vjkoskela:
@BrandonArp how's this look?

src/test/java/com/arpnetworking/tsdcore/sinks/UnitMappingSinkTest.java
```diff
@@ -0,0 +1,83 @@
...
+ */
+public class UnitMappingSinkTest {
+
+    @Before
+    public void setUp() {
+        MockitoAnnotations.initMocks(this);
+    }
+
+    @Test
+    public void testMapping() {
```
BrandonArp: Gimme a couple more tests. Assert on the value of the mapped quantity and show me a "conversion" with no mapping specified.  Easy coverage and could prevent a stupid backwards conversions.

vjkoskela:
@BrandonArp updated.

PR #44, opened 2016-11-28
Docker Image Improvements & Http Routes Entensibility

vjkoskela:
Improvemnets to docker image: i) jvm arguments; ii) logback configuration, iii) updated base image (112 is not yet available). Also includes optional http routes extension.

### DIFF
--- a/src/main/docker/Dockerfile
+++ b/src/main/docker/Dockerfile
@@ -12,23 +12,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM java:8u92-jre-alpine
+FROM openjdk:8u111-jre-alpine
 
 MAINTAINER arpnetworking
 
 EXPOSE 7090
 WORKDIR /opt/mad
 
-# Logging configuration
-ENV LOGBACK_FILE="/opt/mad/config/logback.xml"
-ENV LOGBACK_CONFIG="-Dlogback.configurationFile=${LOGBACK_FILE}"
-
-# JVM configuration
-ENV ADDITIONAL_JAVA_OPTS=""
-
-# MAD configuration
-ENV ADDITIONAL_APP_PARAMS=""
+# Configuration
+ENV JVM_XMS="64m"
+ENV JVM_XMX="1024m"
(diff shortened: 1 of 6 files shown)


src/main/docker/Dockerfile
```diff
@@ -12,22 +12,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM java:8u92-jre-alpine
+FROM java:8u111-jre-alpine
```
BrandonArp: The "java" base image is deprecated.  Please use the opkenjdk base image instead.

src/main/java/com/arpnetworking/http/Routes.java
```diff
@@ -80,16 +80,19 @@
...
      * @param statusPath The path for the status.
+     * @param supplementalRoutes Optional supplemental routes.
      */
     public Routes(
             final ActorSystem actorSystem,
             final MetricsFactory metricsFactory,
             final String healthCheckPath,
-            final String statusPath) {
+            final String statusPath,
+            final Optional<SupplementalRoutes> supplementalRoutes) {
```
BrandonArp: How about making this a list?  Having a single SupplementalRoute object doesn't allow for easy composition.

src/main/docker/Dockerfile
vjkoskela: Dang. Will do.

src/main/java/com/arpnetworking/http/Routes.java
vjkoskela: I'm not sure I see the point in providing it at this level. This provides an entry point to hook into the http router, if the customizer wants to federate the routes, they can do so with whatever design pattern (e.g. configuration, injection, scannotation, etc.) that makes sense in their customization. 

Providing multiple supplements could be useful if we or a second-party were to add a supplement and then a third-party wanted to add another. I think that might be overcomplicating this a bit.

vjkoskela:
Applied changes as discussed. Had to lower the coverage as the code I added is minor and the affected classes already have no tests. Other areas added recently have pushed us down to the limit; and those areas should receive some testing.

PR #49, opened 2017-01-18
update init script, fix config file reference

BrandonArp:


rpm/fs/etc/rc.d/init.d/mad
```diff
@@ -17,15 +17,34 @@ user="mad"
...
+export CONFIG_FILE="/opt/mad/config/config.hocon"
 export JAVA_HOME="$(dirname $(dirname $(readlink -f $(which java))))"
-export JAVA_OPTS="$logging_config"
+export JVM_XMS="64m"
+export JVM_XMX="1024m"
+export LOGGING_CONFIG_FILE="/opt/mad/config/logback.xml"
 
 [ -e /etc/sysconfig/$prog ] && . /etc/sysconfig/$prog
 
+logging_config="-Dlogback.configurationFile=${LOGGING_CONFIG_FILE}"
```
vjkoskela: Why create a variable here? It's only used inside the if-statement on line 32.

rpm/fs/etc/rc.d/init.d/mad
```diff
@@ -17,15 +17,34 @@ user="mad"
...
+if [ -z "$JAVA_OPTS" ]; then
```
vjkoskela: Just curious; do we (by which I mean you) have a use case where you override JAVA_OPTS?

rpm/fs/etc/rc.d/init.d/mad
```diff
@@ -17,15 +17,34 @@ user="mad"
...
-config="/opt/mad/config/config.json"
-logging_config="-Dlogback.configurationFile=/opt/mad/config/logback.xml"
 exec="/opt/mad/bin/mad"
 pid_file="/var/run/mad/mad.pid"
+export CONFIG_FILE="/opt/mad/config/config.hocon"
 export JAVA_HOME="$(dirname $(dirname $(readlink -f $(which java))))"
-export JAVA_OPTS="$logging_config"
+export JVM_XMS="64m"
+export JVM_XMX="1024m"
+export LOGGING_CONFIG_FILE="/opt/mad/config/logback.xml"
```
vjkoskela: I would prefer we matched the Dockerfile's variables where possible; here it would be ```LOGBACK_CONFIG``` instead of just the file name.

vjkoskela: ^ and above it would be ```APP_PARAMS``` instead of just the config file.

vjkoskela:
Minor feedback. Let me know if you want to discuss any of the open questions first.

BrandonArp:
updated to mirror the docker container

rpm/fs/etc/rc.d/init.d/mad
```diff
@@ -41,13 +63,13 @@ is_running() {
 }
 
 start() {
-    [ -f $config ] || exit 6
+    [ -f $CONFIG_FILE ] || exit 6
```
vjkoskela: You should use `${CONFIG_FILE}`. Separate issue completely, it would be nice to spruce up some of the shell conventions here and ideally make it posix compliant. There is a utility called shellcheck that does a pretty good lint for this kind of stuff.

vjkoskela:
Single comment, and a TODO  for the future. Ship it.

PR #52, opened 2017-02-23
Upgrade Akka version.

vjkoskela:


pom.xml
```diff
@@ -71,8 +71,8 @@
 
   <properties>
     <!--Dependency versions-->
-    <akka.version>2.4.2</akka.version>
-    <akka.http.version>2.4.2</akka.http.version>
+    <akka.version>2.4.11.1</akka.version>
```
BrandonArp: Perhaps I'm crazy here, but this doesn't seem like a valid Akka version.  It also isn't the latest (2.4.17).  What's the deal?

vjkoskela: I didn't want to jump to Akka 2.5, but you are correct the versioning is very strange. I picked the latest 2.4.x release which was 2.4.11.1 in January 2017. 

http://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22com.typesafe.akka%22%20AND%20a%3A%22akka-actor_2.11%22

http://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22com.typesafe.akka%22%20AND%20a%3A%22akka-http-core_2.11%22

There is a 2.4.17 that is newer, sorry, I matched the core to the http version. I guess they can be safely unlinked. I just saw both had a 2.4.11.1 and assumed they needed to be linked.

PR #56, opened 2017-03-11
Add a Guice shutdown listener

BrandonArp:


vjkoskela:
Looks good to me; what brought this on?

PR #57, opened 2017-03-11
Use periodic metrics from the incubator extra package

BrandonArp:
This needs to have the version change from -SNAPSHOT.

### DIFF
--- a/src/main/java/com/arpnetworking/http/Routes.java
+++ b/src/main/java/com/arpnetworking/http/Routes.java
@@ -38,12 +38,10 @@ import akka.stream.javadsl.Sink;
 import akka.stream.javadsl.Source;
 import akka.util.ByteString;
 import akka.util.Timeout;
-import com.arpnetworking.metrics.Counter;
-import com.arpnetworking.metrics.Metrics;
-import com.arpnetworking.metrics.MetricsFactory;
-import com.arpnetworking.metrics.Timer;
+import com.arpnetworking.metrics.Units;
 import com.arpnetworking.metrics.common.sources.ClientHttpSourceV1;
 import com.arpnetworking.metrics.common.sources.CollectdHttpSourceV1;
+import com.arpnetworking.metrics.incubator.PeriodicMetrics;
 import com.arpnetworking.metrics.mad.actors.Status;
 import com.arpnetworking.metrics.proxy.actors.Connection;
 import com.arpnetworking.metrics.proxy.models.messages.Connect;
@@ -54,6 +52,7 @@ import com.arpnetworking.steno.LogBuilder;
 import com.arpnetworking.steno.Logger;
 import com.arpnetworking.steno.LoggerFactory;
 import com.google.common.base.Charsets;
+import com.google.common.base.Stopwatch;
 import com.google.common.io.Resources;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import scala.compat.java8.FutureConverters;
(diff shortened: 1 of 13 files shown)


vjkoskela:
Build failed

src/main/java/com/arpnetworking/http/Routes.java
```diff
@@ -78,19 +77,19 @@
...
-     * @param metricsFactory Instance of <code>MetricsFactory</code>.
+     * @param metrics Instance of <code>PeriodicMetrics</code>.
      * @param healthCheckPath The path for the health check.
      * @param statusPath The path for the status.
      * @param supplementalRoutes List of supplemental routes in priority order.
      */
     public Routes(
             final ActorSystem actorSystem,
-            final MetricsFactory metricsFactory,
+            final PeriodicMetrics metrics,
```
vjkoskela: Should it still take the MF which can be used to create the PM?

src/main/java/com/arpnetworking/http/Routes.java
```diff
@@ -111,10 +110,8 @@ public Routes(
      */
     @Override
     public CompletionStage<HttpResponse> apply(final HttpRequest request) {
-        final Metrics metrics = _metricsFactory.create();
-        final Timer timer = metrics.createTimer(createMetricName(request, REQUEST_METRIC));
-        final Counter bodySize = metrics.createCounter(createMetricName(request, BODY_SIZE_METRIC));
-        bodySize.increment(request.entity().getContentLengthOption().orElse(0L));
+        final Stopwatch requestTimer = Stopwatch.createStarted();
```
vjkoskela: Would like to hear about StopWatch versus our Timer and whether we should have a StopWatch if this is the pattern.

src/main/java/com/arpnetworking/metrics/mad/Main.java
```diff
@@ -480,18 +492,41 @@ public void run() {
...
+                    .setMetricsFactory(metricsFactory)
+                    .build();
+            final ScheduledExecutorService executor = Executors.newSingleThreadScheduledExecutor(
+                    r -> new Thread(r, "PeriodicMetricsCloser"));
+            executor.scheduleAtFixedRate(periodicMetrics, 500, 500, TimeUnit.MILLISECONDS);
+
+            // Register to shutdown the executor when the Guice stack is shutdown.
+            lifecycle.registerShutdown(() -> {
+                executor.shutdown();
+                return CompletableFuture.completedFuture(null);
```
vjkoskela: Do we have a variant where this is internalized into the PM?

src/main/java/com/arpnetworking/http/Routes.java
vjkoskela: ^ Based on discussion should take PM because PM will be created from MF; note, that PM interface will be renamed to M (eventually, possibly only in ISM).

src/main/java/com/arpnetworking/http/Routes.java
vjkoskela: Discussion: we decided that PM will be stateful and include our Timer (eventually; possible only in ISM) so the question at hand is Guava vs ARP-M StopWatch.

src/main/java/com/arpnetworking/metrics/mad/Main.java
vjkoskela: As discussed.

vjkoskela:
Seem reasonable to me; just setting us up for some migration work to ISM. I think that's a reasonable trade-off given our need for PM in MAD.

PR #59, opened 2017-04-10
Upgrades to stablize tests and fixes based on the upgrades.

vjkoskela:


vjkoskela:
So on master the dependency tree includes "guava" in two places:

```
[INFO] +- com.google.guava:guava:jar:18.0:compile
[INFO] +- com.fasterxml.jackson.datatype:jackson-datatype-guava:jar:2.7.3:runtim
```

On this branch it's the same but with new versions:

```
[INFO] +- com.google.guava:guava:jar:21.0:compile
[INFO] +- com.fasterxml.jackson.datatype:jackson-datatype-guava:jar:2.7.3:runtime
```

My point, as per your concern, is that there is no transitive dependency on 18.0 being replaced with 21.0 that could cause runtime issues (since there is only one dependency on guava). It's possible there is a `<provided>` dependency on Guava - as there is from commons - but in commons we have already upgraded to 21.0.

@BrandonArp let me know if this is enough to merge this. Thanks!

PR #65, opened 2017-11-01
Added 'bin' tgz artifact.

ddimensia:
I added a new assembly manifest that simply takes the appassembler output and creates a tgz that is attached as a maven artifact.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -305,7 +305,7 @@
           </dependency>
         </dependencies>
       </plugin>
-      <!-- plugin>
+      <plugin>
         <groupId>io.fabric8</groupId>
         <artifactId>docker-maven-plugin</artifactId>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -231,6 +231,21 @@
...
+            <phase>package</phase>
+            <goals>
+              <goal>single</goal>
+            </goals>
+            <configuration>
+              <appendAssemblyId>true</appendAssemblyId>
+              <attach>true</attach>
+              <tarLongFileMode>posix</tarLongFileMode>
+              <descriptors>
+                <descripto>src/main/assembly/bin.xml</descripto>
```
BrandonArp: descriptor

ddimensia: Hmm, I think that tag label must be arbitrary then, given the build succeeded and attached the artifact.  Maven?!?

PR #67, opened 2017-12-02
Add support for a Graphite plaintext source.

vjkoskela:
It's still missing the influx tag format parsing. I'll do that as a separate merge request because a good chunk of adding the graphite source was fixing the underlying tcp line source from the earlier telegraf pull request.


src/main/java/com/arpnetworking/metrics/common/sources/TcpLineSource.java
```diff
@@ -0,0 +1,245 @@
...
+                        .log();
+            }
+        }
+
+        private void processData(final ByteString data) {
+            // Process datagrams line by line buffering any partial lines
+            int indexStart = 0;
+            int indexEnd = data.indexOf('\n');
+
+            // It is possible for there to be more than one line in a datagram
```
BrandonArp: nit: These aren't datagrams.  They're an OS buffer that represents the contents of one or more packets.

src/main/java/com/arpnetworking/metrics/mad/parsers/GraphitePlaintextToRecordParser.java
```diff
@@ -0,0 +1,262 @@
...
+ * - https://github.com/irmen/Pyrolite
+ *
+ * @author Ville Koskela (ville dot koskela at inscopemetrics dot com)
+ */
+public final class GraphitePlaintextToRecordParser implements Parser<List<Record>, ByteBuffer> {
+
+    /**
+     * Parses a graphite plaintext datagram.
+     *
+     * @param datagram a datagram
```
BrandonArp: nit: Not a datagram (see above).

src/main/java/com/arpnetworking/metrics/common/sources/TcpLineSource.java
```diff
@@ -0,0 +1,245 @@
...
+                        .addData("remoteAddress", _remoteAddress.getAddress().getHostAddress())
+                        .addData("remotePort", _remoteAddress.getPort())
+                        .log();
+            }
+        }
+
+        private void processData(final ByteString data) {
+            // Process datagrams line by line buffering any partial lines
+            int indexStart = 0;
+            int indexEnd = data.indexOf('\n');
```
BrandonArp: Your Graphite protocol parser (at least) handles multiple lines.  If you want to make that a requirement of all the parsers, you could use data.lastIndexOf('\n') and get better performance.  Ideally, the buffer will end on a newline character.

src/main/java/com/arpnetworking/metrics/mad/parsers/GraphitePlaintextToRecordParser.java
```diff
@@ -0,0 +1,262 @@
...
+            if (_parseCarbonTags) {
+                // IMPORTANT: Subsequently processed tags can overwrite these
+                final String[] parts = name.split(";");
+                name = parts[0];
+                for (int i = 1; i < parts.length; ++i) {
+                    final String[] subparts = parts[i].split("=");
+                    if (subparts.length == 2 && !subparts[0].isEmpty() && !subparts[1].isEmpty()) {
+                        dimensions.put(subparts[0], subparts[1]);
+                    }
+                    // TODO(ville): should we rate limit log that we potentially dropped something?
```
BrandonArp: I don't think we should fail the whole bytebuffer for invalid data.  If we want to keep the parser processing multiple records, I think it should discard invalid lines, but still process valid lines.

src/main/java/com/arpnetworking/metrics/mad/parsers/GraphitePlaintextToRecordParser.java
```diff
@@ -0,0 +1,262 @@
...
+ *
+ * In addition to the specified features, this parser supports multiple
+ * datagrams separated by newlines.
+ *
+ * The source also allows you to add dimensions in three ways:
+ *
+ * 1) Specifying global tags programmatically.
+ * 2) Parsing tags from the metric name in Carbon format.
+ *     See: http://graphite.readthedocs.io/en/latest/tags.html
+ * 3) Parsing tags from the metric name in InfluxDb format.
```
BrandonArp: Maybe I missed something, but I didn't see this in the code.

src/main/java/com/arpnetworking/metrics/mad/parsers/GraphitePlaintextToRecordParser.java
```diff
@@ -0,0 +1,262 @@
...
+         */
+        public Builder setParseCarbonTags(final Boolean value) {
+            _parseCarbonTags = value;
+            return this;
+        }
+
+        @NotNull
+        private ImmutableMap<String, String> _globalTags = ImmutableMap.of();
+        @NotNull
+        private Boolean _parseCarbonTags = false;
```
BrandonArp: Is there a reason to not do this by default?

BrandonArp:
Overall looks solid.  The influxdb style parsing would be nice, but not required for me.  I can do with a mapper and global tags for the moment.

src/main/java/com/arpnetworking/metrics/mad/parsers/GraphitePlaintextToRecordParser.java
vjkoskela: It's not part of the Graphite spec per say. It's someone's (Carbon's) extension to the schema. While it seems benign (not to mention useful -- way better than the influx extension imo) it is not a standard feature of the protocol and so it's off.

src/main/java/com/arpnetworking/metrics/mad/parsers/GraphitePlaintextToRecordParser.java
vjkoskela: I agree. However, this is a large discussion imo and affects more sinks than just this one. I would like to discuss with you and make changes separately. The goal here was to raise the issue while achieving parity with the current design.

src/main/java/com/arpnetworking/metrics/mad/parsers/GraphitePlaintextToRecordParser.java
vjkoskela: As per MR comments this is not yet implemented. I will update the comment.

src/main/java/com/arpnetworking/metrics/mad/parsers/GraphitePlaintextToRecordParser.java
vjkoskela: Fixed. See above for details.

src/main/java/com/arpnetworking/metrics/common/sources/TcpLineSource.java
vjkoskela: Fixed.

src/main/java/com/arpnetworking/metrics/common/sources/TcpLineSource.java
vjkoskela: Okay, so I think the parser should be responsible for just a single record. So I've moved the line handling into TcpLineSource (as the name would suggest) and removed it from the GraphitePlaintextToRecordParser (it was never in the Telegraf one). 

This will also lead into the discussion about partial failure. That failure will not occur in the parser anymore but in the source. So I think this leads to more consistent semantics where the parser should always throw and the source can skip individual records without discarding everything. 

There are still questions of soft failure, such as with tags in Graphite, and what should be done with records that could be partially recovered. I expect we'll need to address those case by case.

src/main/java/com/arpnetworking/metrics/mad/parsers/GraphitePlaintextToRecordParser.java
BrandonArp: The link specified in the comments is actually documentation for Graphite.  Carbon is the name of the daemon that receives the data and writes it to the (whisper) database.  I'm OK with not making it default, but I think we're safe to do so.  It's a super new feature, but is part of core graphite now.

PR #69, opened 2017-12-08
Add periodic status counts to http server

vjkoskela:


BrandonArp:
Why change the calls from .equals to Objects.equals?

vjkoskela:
We've been using these somewhat interchangeably throughout the code. The `Objects.equals` is null safe, and while we mark our code to lean heavily towards non-null values the fact as our friend Gil would point out is that any of these can still be null (compile-time null safety is not provided). Granted we can infer null safety for many `.equals` calls by swapping the target and argument. If you want to have a convention either way I'm fine, I can see about doing a larger conversion if consistency is your concern.

PR #71, opened 2017-12-12
Adds P75 Statistic to list of available percentile statistics.

ddimensia:


src/main/java/com/arpnetworking/tsdcore/statistics/TP75Statistic.java
```diff
@@ -0,0 +1,33 @@
+/**
+ * Copyright 2014 Brandon Arp
```
BrandonArp: Probably not

PR #74, opened 2017-12-21
Reduce unnecessary MetricReport instance creations.

vjkoskela:


BrandonArp:
Very nice.  I bet this will help the memory churn a lot!

PR #83, opened 2018-01-18
Add host name resolver to Vertx sink.

vjkoskela:
This change allows MAD to connect to CAGG with a custom service discovery layer resolving the configured host "address". By default no resolution is performed and the configured address is used as-is. Since CAGG periodically disconnects MAD clients this should rotate and load balance nicely across a discoverable pool of CAGG hosts.

FYI. @ddimensia

### DIFF
--- a/src/main/java/com/arpnetworking/tsdcore/sinks/VertxSink.java
+++ b/src/main/java/com/arpnetworking/tsdcore/sinks/VertxSink.java
@@ -19,6 +19,8 @@ import com.arpnetworking.logback.annotations.LogValue;
 import com.arpnetworking.steno.LogValueMapFactory;
 import com.arpnetworking.steno.Logger;
 import com.arpnetworking.steno.LoggerFactory;
+import com.arpnetworking.utility.DefaultHostNameResolver;
+import com.arpnetworking.utility.HostNameResolver;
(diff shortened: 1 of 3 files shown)


BrandonArp:
We should talk about how to configure these for use by Guice injection as well.  Think about a case where you want to use a class that looks up EC2 hosts with the AWS API.  You probably want a single client instantiated and you probably don't want to copy the credentials around to multiple locations.

vjkoskela:
I'm not sure I agree. At the moment the only need to copy is if you have multiple pipelines with the same sink. We discussed an alternate configuration scheme where that would not be necessary. Going global of course has the opposite problem where you want separate configurations but can't have them (unless we do some kind of named binding via config). Creating yet another "scope" of configurations adds a lot of complexity. Expanding on our earlier conversation, I we could consider something like this:

```
sources/A.cfg
sources/B.cfg
sinks/X.cfg
sinks/Y.cfg
sinks/Z.cfg
pipelines/Q.cfg
pipelines/R.cfg
```

And then the pipeline configuration can glue the sources and sinks together by name. This would remove any duplication, allow for re-use, while still retaining the ability to address hairy situations where you want some data going only to select destinations or aggregated in different ways.

PR #84, opened 2018-01-18
explicit 500 error on exception, clean up some javadsl stuff

BrandonArp:


src/main/java/com/arpnetworking/http/Routes.java
```diff
@@ -104,7 +103,11 @@ public Routes(
      */
     public Flow<HttpRequest, HttpResponse, NotUsed> flow() {
         return Flow.<HttpRequest>create()
-                .mapAsync(1, this);
+                .mapAsync(1, this)
+                .recover(
+                        new PFBuilder<Throwable, HttpResponse>()
+                                .match(Exception.class, e -> HttpResponse.create().withStatus(StatusCodes.INTERNAL_SERVER_ERROR))
```
vjkoskela: So on what was line 224, there is already an exceptionally in the flow which _may_ map everything to 404. I also don't see where parse exception is caught and mapped to 400. I'm not convinced this works in all cases.

src/main/java/com/arpnetworking/http/Routes.java
```diff
@@ -208,17 +212,28 @@ public Routes(
...
                     ref.tell(new RequestReply(request, response), ActorRef.noSender());
                     return response;
                 })
                 // We return 404 here since actor startup is controlled by config and
                 // the actors may not be running.
-                .exceptionally(err -> HttpResponse.create().withStatus(404));
+                .exceptionally(err -> {
+                    final Throwable cause = err.getCause();
+                    if (cause instanceof ActorNotFound) {
+                        return HttpResponse.create().withStatus(404);
```
vjkoskela: Use the enum like above please.

src/main/java/com/arpnetworking/http/Routes.java
```diff
@@ -208,17 +212,28 @@ public Routes(
...
+                    }
+                    LOGGER.error()
+                            .setMessage("Unhandled exception when looking up actor for http request routing")
+                            .addData("actorName", actorName)
+                            .setThrowable(cause)
+                            .log();
+                    return HttpResponse.create().withStatus(500);
```
vjkoskela: Same here.

vjkoskela:
Just minor change from literals to constants; otherwise looks good. Ship it!

PR #85, opened 2018-02-22
Migration histogram to use fastutil's primitive sorted map.

vjkoskela:
mad-treemap-to-fastutil-histogram

Green = New Version
Red = Existing Version (after all updates in this PR except Fastutil tree implementation change)

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,11 +39,18 @@ In the installation's *bin* directory there are scripts to start Metrics Aggrega
 
 #### Logging
 
-To customize logging you may provide a [LogBack](http://logback.qos.ch/) configuration file.  To use a custom logging configuration you need to define and export an environment variable before executing *mad*:
+To customize logging you may provide a [LogBack](http://logback.qos.ch/) configuration file. The project ships with `logback.xml` which
+writes logs to rotated files and with `logback-console.xml` which writes logs to STDOUT.
 
-    export JAVA_OPTS="-Dlogback.configurationFile=/usr/local/lib/metrics-aggregator-daemon/config/logger.xml"
+Outside of Docker, set the `JAVA_OPTS` environment variable to configure logging:
+
+    export JAVA_OPTS="-Dlogback.configurationFile=/usr/local/lib/metrics-aggregator-daemon/config/logback-console.xml"
 
 Where */usr/local/lib/metrics-aggregator-daemon/config/logger.xml* is the path to your logging configuration file.
+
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -139,38 +146,156 @@ Each of the pipeline configuration files should be placed in the *pipelinesDirec
...
+    "^memcached/memcached_ops/([^/]+)(/value)?": ["memcached/operations/$1"],
+    "^memcached/percent/([^/]+)(/value)?": ["memcached/$1"],
+    "^memcached/ps_count/.*": [],
+    "^memcached/ps_cputime/.*": [],
+    "^uptime/uptime(/value)?": ["uptime/value"]
+  },
+  "source": {
+    type="com.arpnetworking.metrics.common.sources.CollectdHttpSourceV1"
+    actorName="collectd-http-source"
+    name="telegraftcp_source"
```
BrandonArp: Misnamed here

config/pipelines/pipeline.conf
```diff
@@ -0,0 +1,139 @@
...
+  {
+    type="com.arpnetworking.tsdcore.sinks.PeriodFilteringSink"
+    name="telemetry_period_filtering_sink"
+    excludeGreaterThan="PT1S"
+    sink={
+      type="com.arpnetworking.tsdcore.sinks.TelemetrySink"
+      name="telemetry_sink"
+    }
+  }
+]
```
BrandonArp: Newline

vjkoskela:
mad-treemap-to-fastutil-histogram-2

Green = New Version
Red = Existing Version (after all updates in this PR except Fastutil tree implementation change)

In this test, both JVMs were given XMS=XMX=1024m. With no GC pressure (e.g. preallocated large heap) the new FastUtil implementation (green) grows much more slowly than the existing implementation (red). Consequently, it will collect garbage far more infrequently.

Therefore, if you set a high XMS you can reduce GC (and CPU) at the cost of memory consumption. If you set a low XMS you will increase GC (and CPU) but save memory. The same Telegraf workload used about 250Mb in the existing implementation and only about 50Mb in the new implementation.

A few more improvements like this and it's money.

vjkoskela:
Most of the independent stuff was pulled into #86.

README.md
vjkoskela: Fixed.

config/pipelines/pipeline.conf
vjkoskela: Fixed.

PR #88, opened 2018-02-23
Remove commons io usage and dependency.

vjkoskela:


BrandonArp:
I don't understand what most of this PR does. I understand removing the commons.io usage, but what is the rest?

src/main/java/com/arpnetworking/metrics/common/tailer/StatefulTailer.java
```diff
@@ -68,11 +67,14 @@ public void run() {
                         .setThrowable(throwable)
                         .log());
 
-        try {
-            fileLoop();
-        } finally {
-            IOUtils.closeQuietly(_positionStore);
-            IOUtils.closeQuietly(_lineBuffer);
+        try (ByteArrayOutputStream lineBuffer = new ByteArrayOutputStream(INITIAL_BUFFER_SIZE)) {
```
BrandonArp: Why are you creating a new lineBuffer here? Is it just to make sure it gets closed?

vjkoskela: So this code has some serious bugs in it (by my read).

First, it is meant to allow continuation after an exception. If you look at `fileLoop` that method exits when there is an exception but allows the owner to handle it and effectively decide whether to stop or continue (by not calling stop). However, since that method exits it returns here and exits the entire runnable! Hence the introduction of the `while(_isRunning)` loop here as well. 

Second, the line buffer should not be left closed, or rather needs to be re-opened if it's closed. Since the file loop will resume from the last check-point we can safely discard the current buffer thereby ensuring it is open. The easiest and most correct way to manage the buffer is with try-with-resources.

** BUG: I need to move the try-catch inside the while-loop.

Finally, the `positionStore` is closed here, and I have no idea why since this class does not own the instance and it must remain open until the caller decides to stop this instance. Therefore, the lifecycle of that closeable is moved from here to the FileSource.

PR #89, opened 2018-02-23
Switch a couple of classes from slf4j to logback-steno logging.

vjkoskela:


src/main/java/com/arpnetworking/tsdcore/statistics/StatisticFactory.java
```diff
@@ -15,14 +15,14 @@
  */
 package com.arpnetworking.tsdcore.statistics;
 
+import com.arpnetworking.steno.Logger;
+import com.arpnetworking.steno.LoggerFactory;
```
BrandonArp: You need to change the LOGGER calls in this class for them to actually log anything.

vjkoskela: The LogbackSteno logger is backwards compatible. This should work as-is. Although it would be better if there were enhance log statements.

PR #90, opened 2018-02-27
add exponential backoff for statefultailer when the file doesnt exist

BrandonArp:


src/main/java/com/arpnetworking/utility/TimerTrigger.java
```diff
@@ -36,10 +36,19 @@ public TimerTrigger(final Duration duration) {
...
 
     @Override
-    public void waitOnTrigger() throws InterruptedException {
+    public void waitOnReadTrigger() throws InterruptedException {
         Thread.sleep(_duration.getMillis());
     }
 
+    @Override
+    public void waitOnFileNotFoundTrigger(final int attempt) throws InterruptedException {
+        final int maxDuration = (int) ((((Math.random()  //randomize
```
vjkoskela: This I hard to read with the online comments. Can you reformat?

PR #92, opened 2018-03-02
Mapping the dimensions

BrandonArp:
Inject and extract dimensions in the MappingSource


src/main/java/com/arpnetworking/metrics/mad/sources/MappingSource.java
```diff
@@ -116,47 +119,68 @@ public void notify(final Observable observable, final Object event) {
...
                 boolean found = false;
                 for (final Map.Entry<Pattern, List<String>> findAndReplace : _findAndReplace.entrySet()) {
                     final Matcher matcher = findAndReplace.getKey().matcher(metric.getKey());
                     if (matcher.find()) {
                         for (final String replacement : findAndReplace.getValue()) {
-                            merge(metric.getValue(), matcher.replaceAll(replacement), mergedMetrics);
+                            final String replacedString =
+                                    RegexAndMapReplacer.replaceAll(matcher, metric.getKey(), replacement, record.getDimensions());
+
+                            final int tagsStart = replacedString.indexOf(';');
```
vjkoskela: Extracting tags should be configurable based on the replacement.  For backwards compatibility default should probably be not to extract; but I imagine you would prefer the other.

src/main/java/com/arpnetworking/metrics/mad/sources/MappingSource.java
```diff
@@ -116,47 +119,68 @@ public void notify(final Observable observable, final Object event) {
...
+            final Map<ImmutableMap<String, String>, Map<String, MergingMetric>> mergedMetrics = Maps.newHashMap();
             for (final Map.Entry<String, ? extends Metric> metric : record.getMetrics().entrySet()) {
                 boolean found = false;
                 for (final Map.Entry<Pattern, List<String>> findAndReplace : _findAndReplace.entrySet()) {
                     final Matcher matcher = findAndReplace.getKey().matcher(metric.getKey());
                     if (matcher.find()) {
                         for (final String replacement : findAndReplace.getValue()) {
-                            merge(metric.getValue(), matcher.replaceAll(replacement), mergedMetrics);
+                            final String replacedString =
+                                    RegexAndMapReplacer.replaceAll(matcher, metric.getKey(), replacement, record.getDimensions());
```
vjkoskela: If the tag is used in the metric we should be able to configure to remove it from the dimension set. Default should probably be to remove.

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
```diff
@@ -0,0 +1,165 @@
...
+    /**
+     * Replaces all instances of $n (where n is 0-9) with regex match groups, ${var} with regex capture group or variable (from map) 'var'.
+     *
+     * @param matcher matcher to use
+     * @param input input string to match against
+     * @param replace replacement string
+     * @param variables map of variables to include
+     * @return a string with replacement tokens replaced
+     */
+    public static String replaceAll(final Matcher matcher, final String input, final String replace, final Map<String, String> variables) {
```
vjkoskela: Should `variables` be an `ImmutableMap`? -- this also looks long enough to be chopped.

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
```diff
@@ -0,0 +1,165 @@
...
+
+        matcher.reset();
```
vjkoskela: I'm not nuts about the state sharing between this method and the caller.

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
```diff
@@ -0,0 +1,165 @@
...
+        boolean found = matcher.find();
+        if (found) {
+            final StringBuilder builder = new StringBuilder();
+            int lastMatchedIndex = 0;
+            do {
+                builder.append(input.substring(lastMatchedIndex, matcher.start()));
+                lastMatchedIndex = matcher.end();
+                appendReplacement(matcher, replace, builder, variables);
+                found = matcher.find();
+            } while (found);
```
vjkoskela: The matcher is for the metric pattern; what matches are you iterating over here? I am a little concerned you've combined two concepts.

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
```diff
@@ -0,0 +1,165 @@
...
+            // Append left-over string after the matches
+            if (lastMatchedIndex < input.length() - 1) {
+                builder.append(input.substring(lastMatchedIndex, input.length()));
+            }
+            return builder.toString();
+        }
+        return input;
+    }
+
+    private static void appendReplacement(final Matcher matcher, final String replacement, final StringBuilder replacementBuilder,
```
vjkoskela: chop-all

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
```diff
@@ -0,0 +1,165 @@
...
+import java.util.regex.Matcher;
+
+/**
+ * A regex replacement utility that can also replace tokens not found in the regex.
+ *
+ * @author Brandon Arp (brandon dot arp at inscopemetrics dot com)
+ */
+public final class RegexAndMapReplacer {
+    /**
+     * Replaces all instances of $n (where n is 0-9) with regex match groups, ${var} with regex capture group or variable (from map) 'var'.
```
vjkoskela: Not nuts about `$n` versus `${var}` since these are commonly the same thing. I think it would be cleaner to only support `${...}` and just have precedence rules (you already need them for named capture versus key). I'm guessing that capture groups would take precedence over dimensions (`variables`); but I am not entirely sure. If there is no clear answer or we want to provide flexibility I would recommend namespacing (e.g. `${dimension:key}` and `${capture:name}`)

src/test/java/com/arpnetworking/metrics/mad/sources/MappingSourceTest.java
```diff
@@ -278,6 +281,140 @@ public void testReplaceWithCapture() {
...
+                                        new Quantity.Builder()
+                                                .setValue(1.23d)
+                                                .setUnit(Unit.BYTE)
+                                                .build()))
+                                .build()))
+                .build();
+        Assert.assertTrue(
+                String.format("expected=%s, actual=%s", expectedRecord, actualRecord),
+                UnorderedRecordEquality.equals(expectedRecord, actualRecord));
+    }
```
vjkoskela: Can you refactor some helpers out of the tests?

src/test/java/com/arpnetworking/metrics/mad/sources/MappingSourceTest.java
```diff
@@ -52,7 +53,9 @@ public void setUp() {
                 .setName("MergingSourceTest")
                 .setFindAndReplace(ImmutableMap.of(
                         "foo/([^/]*)/bar", ImmutableList.of("foo/bar"),
-                        "cat/([^/]*)/dog", ImmutableList.of("cat/dog", "cat/dog/$1")))
+                        "cat/([^/]*)/dog", ImmutableList.of("cat/dog", "cat/dog/$1"),
+                        "tagged/([^/]*)/dog", ImmutableList.of("tagged/dog;animal=$1"),
+                        "tagged/([^/]*)/animal", ImmutableList.of("tagged/${animal}/animal")))
```
vjkoskela: Include example with named capture and then one with precedence.

src/test/java/com/arpnetworking/utility/RegexAndMapBenchmarkTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+import java.util.Collections;
+import java.util.regex.Pattern;
+
+/**
+ * Benchmark tests comparing the RegexAndMapReplacer class against the built-in regex replacement.
+ *
+ * @author Brandon Arp (brandon dot arp at inscopemetrics dot com)
+ */
+public class RegexAndMapBenchmarkTest {
```
vjkoskela: - `final`
- Not sure this should be run with regular unit tests; at Q we separate PT into E2E-PT and PT

vjkoskela: But very happy to see this.

src/test/java/com/arpnetworking/utility/RegexAndMapReplacerTest.java
```diff
@@ -0,0 +1,199 @@
...
+import java.util.Collections;
+import java.util.Map;
+import java.util.regex.Pattern;
+
+/**
+ * Tests for the {@link RegexAndMapReplacer} class.
+ *
+ * @author Brandon Arp (brandon dot arp at inscopemetrics dot com)
+ */
+public class RegexAndMapReplacerTest {
```
vjkoskela: - `final`
- What's the coverage this gives us?

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    private static void appendReplacement(final Matcher matcher, final String replacement, final StringBuilder replacementBuilder,
+                                          final Map<String, String> variables) {
+        final StringBuilder tokenBuilder = new StringBuilder();
+        int x = -1;
+        while (x < replacement.length() - 1) {
+            x++;
+            final char c = replacement.charAt(x);
+            if (c == '\\') {
```
vjkoskela: Is `\\` used to escape `$` or should we just use `$$`? Either way, this needs to be documented in the class header.

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
```diff
@@ -0,0 +1,165 @@
...
+        }
+    }
+
+    private static void processEscapedCharacter(final String replacement, final int x, final StringBuilder builder) {
+        if (x >= replacement.length()) {
+            throw new IllegalArgumentException(
+                    String.format("Improper escaping in replacement, must not have trailing '\\' at col %d: %s", x, replacement));
+        }
+        final Character c = replacement.charAt(x);
+        if (c == '\\' || c == '$' || c == '}') {
```
vjkoskela: I believe with most open-close style formats you need to escape both. I understand it's strictly not necessary here, but I'd like to be consistent with say regex and bash.

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
```diff
@@ -0,0 +1,165 @@
...
+            builder.append(c);
+        } else {
+            throw new IllegalArgumentException(
+                    String.format("Improperly escaped '%s' in replacement at col %d: %s", c, x, replacement));
+        }
+    }
+
+    private static int writeReplacementToken(final String replacement, final int offset, final StringBuilder output,
```
vjkoskela: chop

src/test/java/com/arpnetworking/utility/RegexAndMapReplacerTest.java
```diff
@@ -0,0 +1,199 @@
...
+    private void testExpression(final Pattern pattern, final String input, final String replace, final String expected,
+            final Map<String, String> variables) {
+        final String result = RegexAndMapReplacer.replaceAll(pattern.matcher(input), input, replace, variables);
+        Assert.assertEquals(expected, result);
+        try {
+            final String stockResult = pattern.matcher(input).replaceAll(replace);
+            Assert.assertEquals(expected, stockResult);
+        } catch (final IllegalArgumentException ignored) { }
+    }
+}
```
vjkoskela: We should take this opportunity to document the mapping source in the README.md.

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
BrandonArp: Fixex

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
BrandonArp: Sure, why not? Fixed.

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
BrandonArp: As discussed.  It kinda is, but it's also a fairly succinct way of specifying the intent.

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
BrandonArp: Fixed.

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
BrandonArp: $n is already supported by the MappingSink.  It's probably the most common case that people are familiar with and I hesitate to remove it.  There are precedence rules.  And you are correct that the capture group is picked over a dimension of the same name.  Adding namespacing seems like overkill given that you can just change the name of the capture group (since the name isn't used for anything but the replacement).

src/test/java/com/arpnetworking/utility/RegexAndMapBenchmarkTest.java
BrandonArp: Eh, it only takes a few seconds to run, so I'm not too concerned.  Lets talk offline about how to add better perf tests.

src/test/java/com/arpnetworking/utility/RegexAndMapReplacerTest.java
BrandonArp: Fixed.  Coverage is 100%

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
BrandonArp: Documented.  Also note that \\ is just a java-ism for the literal '\'

src/main/java/com/arpnetworking/metrics/mad/sources/MappingSource.java
BrandonArp: We should talk about this.  It's gonna be a bit tricky.

jdk-wrapper.sh
```diff
@@ -40,6 +40,27 @@ safe_command() {
   fi
 }
 
+checksum() {
```
vjkoskela: Thank you.

src/main/java/com/arpnetworking/metrics/mad/sources/TransformingSource.java
```diff
@@ -0,0 +1,429 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(TransformingSource.class);
+    private static final Splitter.MapSplitter TAG_SPLITTER = Splitter.on(';').omitEmptyStrings().trimResults().withKeyValueSeparator('=');
+
+    // NOTE: Package private for testing
+    /* package private */ static final class TransformingObserver implements Observer {
+
+        /* package private */ TransformingObserver(
+                final TransformingSource source,
+                final Map<Pattern, List<String>> findAndReplace,
```
vjkoskela: This should be ImmutableMap; no?

src/main/java/com/arpnetworking/metrics/mad/sources/TransformingSource.java
```diff
@@ -0,0 +1,429 @@
...
+    @Override
+    public String toString() {
+        return toLogValue().toString();
+    }
+
+    private TransformingSource(final Builder builder) {
+        super(builder);
+        _source = builder._source;
+
+        final ImmutableMap.Builder<Pattern, List<String>> findReplaceBuilder =
```
vjkoskela: List => ImmutableList

src/main/java/com/arpnetworking/metrics/mad/sources/TransformingSource.java
vjkoskela: (and ImmutableList)

src/main/java/com/arpnetworking/metrics/mad/sources/TransformingSource.java
```diff
@@ -0,0 +1,429 @@
...
+                LOGGER.error()
+                        .setMessage("Observed unsupported event")
+                        .addData("event", event)
+                        .log();
+                return;
+            }
+
+            // Merge the metrics in the record together
+            final Record record = (Record) event;
+            final Map<ImmutableMap<String, String>, Map<String, MergingMetric>> mergedMetrics = Maps.newHashMap();
```
vjkoskela: Do you think we should eventually push the `Key` model class to this level? If so, can you add a TODO here? (if not, let me know what your thoughts are on it)

src/main/java/com/arpnetworking/metrics/mad/sources/TransformingSource.java
```diff
@@ -0,0 +1,429 @@
...
+            for (final Map.Entry<String, ? extends Metric> metric : record.getMetrics().entrySet()) {
+                boolean found = false;
+                final String metricName = metric.getKey();
+                for (final Map.Entry<Pattern, List<String>> findAndReplace : _findAndReplace.entrySet()) {
+                    final Pattern metricPattern = findAndReplace.getKey();
```
vjkoskela: Can you rename this `findAndReplacePattern`?

The one thing I am not seeing here is how to select which metrics to apply which rules to. The difference between selecting a metrics and transforming it. While I realize that for metric name this can often be expressed with one regex, I suspect there is actually more to this pattern.

Specifically, I believe we'll need to provide "selection" of metrics based not only on name but also on dimension set. For example: contain key, containing key regex or containing key with value or regex of value; I don't see a point in just containing value or regex of value ... but that's probably my overly typed view of the world.

Implementing the full selection is not something you need to do now. However, I would like you to consider it in how the configuration is laid out. I would like to be able to add this without making non-backwards compatible changes to the configuration. To this end I believe we need to support a list of transformations. By default the transformation can match all, and then adding selectors can narrow those down. There is one question which is whether transformations are all applied or only some.

We can default either way, supporting the other can just be a flag away. Although I am concerned about how slippery the slope may be to a graph/tree representation (instead of a list). I would like to hear your thoughts on this one.

src/main/java/com/arpnetworking/metrics/mad/sources/TransformingSource.java
```diff
@@ -0,0 +1,429 @@
...
+                    final Matcher matcher = metricPattern.matcher(metricName);
+                    if (matcher.find()) {
+                        for (final String replacement : findAndReplace.getValue()) {
```
vjkoskela: Can you factor this out into a separate method. Should improve readability and testability.

src/main/java/com/arpnetworking/metrics/mad/sources/TransformingSource.java
```diff
@@ -0,0 +1,429 @@
...
+                            metric.getValue(),
+                            metricName,
+                            mergedMetrics,
+                            getModifiedDimensions(record.getDimensions(), Collections.emptyMap(), ImmutableList.of()));
+                }
+            }
+
+            // Raise the merged record event with this source's observers
+            // NOTE: Do not leak instances of MergingMetric since it is mutable
+            for (Map.Entry<ImmutableMap<String, String>, Map<String, MergingMetric>> entry : mergedMetrics.entrySet()) {
```
vjkoskela: nit: final entry

src/main/java/com/arpnetworking/metrics/mad/sources/TransformingSource.java
```diff
@@ -0,0 +1,429 @@
...
+                                        .setAnnotations(record.getAnnotations())
+                                        .setDimensions(entry.getKey())));
+            }
+        }
+
+        private ImmutableMap<String, String> getModifiedDimensions(
+                final ImmutableMap<String, String> inputDimensions,
+                final Map<String, String> add,
+                final ImmutableList<String> remove) {
+            final Map<String, String> finalTags = Maps.newHashMap();
```
vjkoskela: You should probably initialize the expected size.

vjkoskela: or else use a constructor that accepts `inputDimensions`

src/main/java/com/arpnetworking/metrics/mad/sources/TransformingSource.java
```diff
@@ -0,0 +1,429 @@
...
+            _inject.forEach(
+                    (key, inject) ->
+                            finalTags.compute(key, (k, oldValue) ->
+                                    inject.isReplaceExisting() || oldValue == null ? inject.getValue() : oldValue));
+            finalTags.putAll(add);
+
+            return ImmutableMap.copyOf(finalTags);
+        }
+
+        private void merge(final Metric metric, final String key,
```
vjkoskela: wrap one, wrap all

src/main/java/com/arpnetworking/metrics/mad/sources/TransformingSource.java
```diff
@@ -0,0 +1,429 @@
...
+                return this;
+            }
+
+            /**
+             * Whether to override existing dimension of this name. Optional. Cannot be null. Defaults to true.
+             *
+             * @param value true to replace existing dimension value
+             * @return This instance of {@link Builder}.
+             */
+            public Builder setReplaceExisting(final Boolean value) {
```
vjkoskela: nit: overwrite?

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
```diff
@@ -0,0 +1,216 @@
...
+package com.arpnetworking.utility;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
```
vjkoskela: Please add a TODO to refactor this into commons once there is a second use case for it.

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
vjkoskela: There is no need to maintain compatibility with the mapping sink. I think that there are a few other cases here which may be handy:

`${env:<name>}`
`${prop:<name>}`
`${capture:<name|number>}`
`${dimension:<name>}`

While the simplicity of precedence is appealing, I cannot come up with a consistent scheme that doesn't break down. The namespacing while somewhat more arduous on users is explicit and easy to debug in case of failure (versus precedence which can result in unexpected results).

If you feel strongly another way I'd like to hear your thoughts on this.

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
```diff
@@ -0,0 +1,216 @@
...
+        if (c == '{') {
+            inReplaceBrackets = true;
+            x++;
+            c = replacement.charAt(x);
+        }
+
+        if (inReplaceBrackets) {
+            // Consume until we hit the }
+            while (x < replacement.length() - 1 && c != '}') {
+                if (c == '\\') {
```
vjkoskela: Like regex capture groups, there should be no need to support escaped characters inside replacements unless you want to support dynamic replacements; 

e.g.
`${foo_${bar}}`

This would certainly give insane flexibility.

src/main/java/com/arpnetworking/metrics/mad/sources/TransformingSource.java
BrandonArp: Fixed

src/main/java/com/arpnetworking/metrics/mad/sources/TransformingSource.java
BrandonArp: Fixed

src/main/java/com/arpnetworking/metrics/mad/sources/TransformingSource.java
BrandonArp: Fixed

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
BrandonArp: Allowing escaping does not allow for dynamic replacements like that.  It does allow for "special" characters in the names of the dimension, however. For example, if someone decided to have "{a}" as a dimension name, for whatever reason.  Moreover, it makes for more consistent rules about how to escape (especially important if you have multiple levels of escaping.

src/main/java/com/arpnetworking/metrics/mad/sources/TransformingSource.java
BrandonArp: Fixed

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
BrandonArp: Added the prefix support (env, capture, dimension are supported now).  Precedence is explicit in the ordering of the prefixed variable maps.  Phew, that was a bit of work.

README.md
```diff
@@ -285,6 +285,80 @@ interval="10s"
...
+```
+
+This namespacing prevents the built-in precendence search for a matching variable.
+
+__Note: Any dimensions matched as part of a replacement will be removed from the resulting metric.__
+
+Dimensions can also be injected into a metric.  To do this, add a ';' after the replacement name of the metric
+and proceed to specify key=value pairs (where both the key and value can use variables).  Basically, the
+replacement is processed, the string split on ';' with the first part being the metric name and anything
+after the ';' being parsed as key=value pairs to be added to the metric's dimensions.
```
vjkoskela: Do they replace? What if I don't want them to?

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
```diff
@@ -0,0 +1,296 @@
...
+    /**
+     * Replaces all instances of $n (where n is 0-9) with regex match groups, ${var} with regex capture group or variable (from map) 'var'.
+     *
+     * @param pattern pattern to use
+     * @param input input string to match against
+     * @param replace replacement string
+     * @param variables map of variables to include
+     * @return a string with replacement tokens replaced
+     */
+        public static Replacement replaceAll(
```
vjkoskela: Indentation looks off.

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
```diff
@@ -0,0 +1,296 @@
...
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * A regex replacement utility that can also replace tokens not found in the regex.
+ *
+ * $n where n is a number in the replace string is replaced by the pattern's match group n
+ * ${name} in the replace string is replaced by the pattern's named capture, or by the value of the variable
```
vjkoskela: This documentation doesn't match the documentation in the README. Also, I thought the numeric support was going to be removed altogether. Having the $0 format supported begs the question why not support $foo as well with white-space parsing a la bash. I think it's a path we shouldn't go down.

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
```diff
@@ -0,0 +1,296 @@
...
+        return x - offset - 1;
+    }
+
+    private static String getReplacement(
+            final Matcher matcher,
+            final String replaceToken,
+            final boolean numeric,
+            final LinkedHashMap<String, Map<String, String>> variables,
+            final ImmutableList.Builder<String> variablesUsedBuilder) {
+        if (numeric) {
```
vjkoskela: Even if it's numeric it can still be a dimension or env var name.

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
```diff
@@ -0,0 +1,296 @@
...
+            final int replaceGroup = Integer.parseInt(replaceToken);
+            return matcher.group(replaceGroup);
+        } else {
+            final int prefixSeparatorIndex = replaceToken.indexOf(':');
+            final String prefix;
+            final String variableName;
+            if (prefixSeparatorIndex == replaceToken.length() - 1) {
```
vjkoskela: Do you need to be able to escape the colon?

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
```diff
@@ -0,0 +1,296 @@
...
+
+    private static String replacePrefix(
+            final Matcher matcher,
+            final LinkedHashMap<String, Map<String, String>> variables,
+            final ImmutableList.Builder<String> variablesUsedBuilder,
+            final String prefix,
+            final String variableName) {
+        if (prefix.equals("capture")) {
+            if (isNumeric(variableName)) {
+                final Integer groupIndex = Integer.valueOf(variableName);
```
vjkoskela: Same comment about about numeric being something other than a capture group index and also the function calling this already special cased this.

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
```diff
@@ -0,0 +1,296 @@
...
+            } else {
+                try {
+                    return matcher.group(variableName);
+                } catch (final IllegalArgumentException ignored2) { // No group with this name
+                    return "";
+                }
+            }
+        }
+
+        // Only record variables that are not captures
```
vjkoskela: I think you only need to record variables that are dimensions. However, see my comments on the README about the default to remove.

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
```diff
@@ -0,0 +1,296 @@
...
+        variablesUsedBuilder.add(String.format("%s:%s", prefix, variableName));
+
+        final Map<String, String> variableMap = variables.get(prefix);
```
vjkoskela: Does the variableMap contain the environment variables too?

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    private static String replaceNonPrefix(
+            final Matcher matcher,
+            final LinkedHashMap<String, Map<String, String>> variables,
+            final ImmutableList.Builder<String> variablesUsedBuilder,
+            final String variableName) {
+        // First try the capture group with the name
+        try {
+            return matcher.group(variableName);
```
vjkoskela: If you are intent on keeping the numeric representation, I think that you need to push it down to this level to ensure it's used in the correct precedence order.

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
vjkoskela: ^ This can also throw or be invalid if it overflows.

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
```diff
@@ -0,0 +1,296 @@
...
+        return isNumeric;
+    }
+
+    private RegexAndMapReplacer() { }
+
+    /**
+     * Describes the replacement string and variables used in it's creation.
+     *
+     * The "replacement" field is the resulting string.
+     * The "variablesMatched" field is a list of input variables that were matched, in prefix:variable form
```
vjkoskela: I'm not nuts about maintaining the internal data structures with the prefixes. The reason is that it means parsing and escaping the colons in more than one place. Have you considered just using the prefix to refer to a different input map? (e.g. map of maps where first key is prefix -- for example)

README.md
```diff
@@ -285,6 +285,80 @@ interval="10s"
...
+```bash
+${env:MYVAR}
+```
+```bash
+${dimension:some_dimension}
+```
+
+This namespacing prevents the built-in precendence search for a matching variable.
+
+__Note: Any dimensions matched as part of a replacement will be removed from the resulting metric.__
```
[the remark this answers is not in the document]
BrandonArp: Removal of dimensions is not on a per-metric basis.  The removal applies to all metrics. As for preserving it, it's as simple as "metric/${consumed};consumed=${consumed}".  Basically, add it as a dimension with the same value.

README.md
BrandonArp: If you didn't want them to replace, why would you put them there?

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
BrandonArp: The RegexAndMapReplacer doesn't know about the concept of dimensions. I was hoping to keep this class separate from the implementation details of the source.

README.md
```diff
@@ -285,6 +285,80 @@ interval="10s"
...
+      inject = {
+        foo {
+          value = bar
+          overwriteExisting = false
+        }
+      }
+      remove = [
+        baz
+      ]
+      findAndReplace = {
```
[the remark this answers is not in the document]
BrandonArp: I'll change the names. I don't think the semicolon can be escaped.  I doubt that's something we'll really see, but I can add that functionality if you really feel strongly about it.  Also, should I then be allowing escaping the '='?

I agree that your configuration example is more powerful (in most cases), but it adds significantly more processing (even after parsing is considered).  If you want to add multiple dimensions, you need to run each dimension value through that same replacement, with the same context from the match.  Also, you'll lose the ability to do something like "some/([^/]*/([^/]*/val;${1}=${2}" where you can have dynamic dimension keys.  Moving the inject and remove dimensions into the metric transformation will likely be a significant code change for, IMHO, little value and significantly increased cost of use (configuration, processing time).

README.md
```diff
@@ -285,6 +285,80 @@ interval="10s"
...
+  source {
+    // your wrapped source goes here
+  }
+}
+```
+
+tranformations is a list of TransformationSet objects.  Each TransformationSet has an inject (Map\<String, DimensionInjection\>),
+ remove (List\<String\>) and a findAndReplace (Map\<String, List\<String\>\>).
+
+ A DimensionInjection is just a value and a boolean of whether or not to overwrite existing values.
```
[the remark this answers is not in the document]
BrandonArp: It does.  Documented.

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
BrandonArp: Yeah.... wtf?  Fixed.

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
BrandonArp: Agreed. I didn't consider the case where an environment var, for example, was a number.

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
BrandonArp: That's fair. I'll return that same map structure as gets passed in.

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
BrandonArp: Fixed.

README.md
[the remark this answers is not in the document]
vjkoskela: Sorry, that's a weird scoping; why would the removal apply to all metrics but when the extraction only applies to some?

README.md
```diff
@@ -285,6 +285,80 @@ interval="10s"
...
+  source {
+    // your wrapped source goes here
+  }
+}
+```
+
+tranformations is a list of TransformationSet objects.  Each TransformationSet has an inject (Map\<String, DimensionInjection\>),
+ remove (List\<String\>) and a findAndReplace (Map\<String, List\<String\>\>).
+
+ A DimensionInjection is just a value and a boolean of whether or not to overwrite existing values. Of omitted, overwrite defaults to true.
```
vjkoskela: Type-o: `Of` => `If`

README.md
```diff
@@ -285,6 +285,80 @@ interval="10s"
...
+      }
+    }
+  ]
+  source {
+    // your wrapped source goes here
+  }
+}
+```
+
+tranformations is a list of TransformationSet objects.  Each TransformationSet has an inject (Map\<String, DimensionInjection\>),
```
vjkoskela: t => T

vjkoskela: Instead of escaping, why not put it in ticks for inline code?

README.md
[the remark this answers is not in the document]
vjkoskela: This is unclear and somewhat surprising since the rule only applies to a particular set of metrics based on the regex match.

src/main/java/com/arpnetworking/metrics/mad/sources/TransformingSource.java
```diff
@@ -0,0 +1,490 @@
...
+            }
+
+            // Merge the metrics in the record together
+            final Record record = (Record) event;
+            final Map<Key, Map<String, MergingMetric>> mergedMetrics = Maps.newHashMap();
+            final LinkedHashMap<String, Map<String, String>> variablesMap = new LinkedHashMap<>();
+            variablesMap.put("dimension", record.getDimensions());
+            variablesMap.put("env", System.getenv());
+
+            for (TransformationSet transformation : _transformations) {
```
vjkoskela: I'm not convinced that the transformations should be treated as a single context. As a single context there can be only a single set of dimensions. In the case of say CollectD this is not necessarily the case. Consider cpu extracting core into a dimension; that doesn't apply to memory (were they in the same UOW). I think each transformation needs to create its own UOW from the existing one with the matching metrics and the dimensions that are elected. If we want to join them back (assuming dimension sets match) that's a possible optimization. Thoughts?

vjkoskela: To elaborate:

- There is the incoming UOW
- Metrics matching a transformation are transformed into a new UOW which inherits its dimensions from the parent (post inject and remove) along with any injections/removals from the transform
- Any metrics included in a new UOW are removed from the parent UOW
- Metrics matching multiple UOWs are duplicated across them (subsequent changes adding separate matching rules to transformers can make this more elaborate)

src/main/java/com/arpnetworking/metrics/mad/sources/TransformingSource.java
```diff
@@ -0,0 +1,490 @@
...
+        private Key getModifiedDimensions(
+                final ImmutableMap<String, String> inputDimensions,
+                final Map<String, String> add,
+                final List<String> remove,
+                final TransformationSet transformation) {
+            final Map<String, String> finalTags = Maps.newHashMap(inputDimensions);
+            // Remove the dimensions that we consumed in the replacement
+            remove.forEach(finalTags::remove);
+            transformation.getRemoveDimensions().forEach(finalTags::remove);
+            transformation.getInjectDimensions().forEach(
```
vjkoskela: You should document that inject trumps remove and both of these trump anything extracted in the transform and subsequently removed. However, I'm a little concerned about the way this is done. Specifically, the way these can compound. Depending on your feedback on the independent UOWs from each transformation this may be a discussion we should have.

src/main/java/com/arpnetworking/metrics/mad/sources/TransformingSource.java
```diff
@@ -0,0 +1,490 @@
...
+        }
+
+        private final MetricType _type;
+        private final ImmutableList.Builder<Quantity> _values = ImmutableList.builder();
+    }
+
+    /**
+     * Represents a dimension to inject and whether or not it should overwrite the existing value (if any).
+     */
+    public static final class DimensionInjection {
```
vjkoskela: Needs to be marked `@Loggable`.

src/main/java/com/arpnetworking/metrics/mad/sources/TransformingSource.java
```diff
@@ -0,0 +1,490 @@
...
+            private Boolean _overwrite = true;
+        }
+    }
+
+    /**
+     * Represents a set of transformations to apply.
+     *
+     * @author Brandon Arp (brandon dot arp at inscopemetrics dot com)
+     */
+    public static final class TransformationSet {
```
vjkoskela: Needs to be marked `@Loggable`.

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
```diff
@@ -0,0 +1,294 @@
...
+    private static int writeReplacementToken(
+            final String replacement,
+            final int offset,
+            final StringBuilder output,
+            final Matcher matcher,
+            final LinkedHashMap<String, Map<String, String>> variables,
+            final StringBuilder tokenBuilder,
+            final Map<String, ImmutableList.Builder<String>> variablesUsed) {
+        boolean inReplaceBrackets = false;
+        boolean tokenNumeric = true;
```
vjkoskela: unused

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
```diff
@@ -0,0 +1,294 @@
...
+        final int prefixSeparatorIndex = replaceToken.indexOf(':');
+        final String prefix;
+        final String variableName;
+        if (prefixSeparatorIndex == replaceToken.length() - 1) {
+            throw new IllegalArgumentException(
+                    String.format("found prefix in variable replacement, but no variable name found: '%s'",
+                    replaceToken));
+        }
+        if (prefixSeparatorIndex != -1) {
+            prefix = replaceToken.substring(0, prefixSeparatorIndex);
```
vjkoskela: What happens if I have `${:foo}`; won't that be processed as non-prefixed? Where is prefix validation done?

vjkoskela: (empty prefix is not the same as no prefix)

src/main/java/com/arpnetworking/metrics/mad/sources/TransformingSource.java
BrandonArp: That's essentially what we're doing with the merge map.

src/main/java/com/arpnetworking/metrics/mad/sources/TransformingSource.java
BrandonArp: That's not how this is done. 1) remove the consumed dimensions, 2) remove the explicitly removed dimensions, 3) inject the explicitly injected dimensions (honoring overwrite), 4) inject the extracted dimensions.  Not sure what you think the order should be.

src/main/java/com/arpnetworking/metrics/mad/sources/TransformingSource.java
BrandonArp: Done

src/main/java/com/arpnetworking/metrics/mad/sources/TransformingSource.java
BrandonArp: Done

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
BrandonArp: Removed.

src/main/java/com/arpnetworking/utility/RegexAndMapReplacer.java
BrandonArp: The way these are processed, empty prefix is the same as no prefix.

PR #95, opened 2018-03-24
add a source that injects dimensions

BrandonArp:


src/main/java/com/arpnetworking/metrics/mad/sources/DimensionInjectingSource.java
```diff
@@ -0,0 +1,195 @@
...
+
+        @Override
+        protected Builder self() {
+            return this;
+        }
+
+        @NotNull
+        private Source _source;
+        @NotNull
+        private Map<String, String> _dimensions;
```
vjkoskela: Why isn't this a `ImmutableMap`?

src/main/java/com/arpnetworking/metrics/mad/sources/DimensionInjectingSource.java
```diff
@@ -0,0 +1,195 @@
...
+        @NotNull
+        private Boolean _replaceExisting = true;
```
vjkoskela: Should this be a per dimension setting? I guess you could nest two of these sources; however, that is more data copy. After the work on the mapping dimensions, I've been thinking that we could do this type of injection closer to the source and therefore avoid the copying altogether.

src/test/java/com/arpnetworking/metrics/mad/sources/DimensionInjectingSourceTest.java
```diff
@@ -0,0 +1,198 @@
...
+import java.util.HashMap;
+
+/**
+ * Tests for the {@link DimensionInjectingSource} class.
+ *
+ * @author Brandon Arp (brandon dot arp at inscopemetrics dot com)
+ */
+public class DimensionInjectingSourceTest {
+
+
```
vjkoskela: nit: extra blank line.

src/test/java/com/arpnetworking/test/UnorderedRecordEquality.java
```diff
@@ -63,6 +64,11 @@ public static boolean equals(final Record r1, final Record r2) {
             }
         }
 
+        for (Map.Entry<String, String> entry : r1.getDimensions().entrySet()) {
```
vjkoskela: `final` iterator?

src/main/java/com/arpnetworking/metrics/mad/sources/DimensionInjectingSource.java
BrandonArp: I'd rather not complicate it like that. We haven't even had this source at all. Mixing id's probably unnecessary for now. If there is a use case I can add it.

vjkoskela: Sure we haven't had this because most of the use case have not supported arbitrary dimension aggregation; we've had the equivalent sink though.

src/main/java/com/arpnetworking/metrics/mad/sources/DimensionInjectingSource.java
```diff
@@ -0,0 +1,195 @@
...
+
+    // NOTE: Package private for testing
+    /* package private */ static final class InjectingObserver implements Observer {
+
+        /* package private */ InjectingObserver(
+                final DimensionInjectingSource source,
+                final boolean replace,
+                final ImmutableMap<String, String> dimensions) {
+            _source = source;
+            _replace = replace;
```
vjkoskela: nit: replace to overwrite (or change the docs to match replace)

src/main/java/com/arpnetworking/metrics/mad/sources/DimensionInjectingSource.java
```diff
@@ -0,0 +1,195 @@
...
+
+            // Merge the metrics in the record together
+            final Record record = (Record) event;
+            final Map<String, String> newDimensions;
+            if (_replace) {
+                newDimensions = Maps .newHashMapWithExpectedSize(record.getAnnotations().size() + _dimensions.size());
+                newDimensions.putAll(record.getDimensions());
+                newDimensions.putAll(_dimensions);
+            } else {
+                newDimensions = Maps .newHashMapWithExpectedSize(record.getAnnotations().size() + _dimensions.size());
```
vjkoskela: The initialization  can be moved out of the if-else branch no?

src/main/java/com/arpnetworking/metrics/mad/sources/DimensionInjectingSource.java
```diff
@@ -0,0 +1,195 @@
...
+            return this;
+        }
+
+        /**
+         * Sets the dimensions. Required. Cannot be null.
+         *
+         * @param value The map of dimensions and their values
+         * @return This instance of {@link Builder}.
+         */
+        public Builder setDimensions(final Map<String, String> value) {
```
vjkoskela: Why isn't this an ImmutableMap in the setter?

src/main/java/com/arpnetworking/metrics/mad/sources/DimensionInjectingSource.java
```diff
@@ -0,0 +1,195 @@
...
+            return this;
+        }
+
+        /**
+         * Sets whether or not to replace existing values. Optional. Cannot be null. Defaults to true.
+         *
+         * @param value True to replace values
+         * @return This instance of {@link Builder}.
+         */
+        public Builder setReplaceExisting(final Boolean value) {
```
vjkoskela: nit: overwrite is probably simpler here too

src/test/java/com/arpnetworking/metrics/mad/sources/DimensionInjectingSourceTest.java
vjkoskela: ?

PR #97, opened 2018-05-30
Whitespace/EOF cleanup, build control changes

ddimensia:
There are two commits that contains cleanup that resulted from running a linter against the source three:

99f89a2 Clean up missing EOF and unnecessary whitespace.
c4ee019 Added 'set -e' to cause the scripts to fail on first non-zero...

The final commit enables the ability to exert more control over the build.  It allows for the disabling of the docker plugin via disabling a 'docker' profile.  It also allows for control of the output directory and output filename of the tgz file.

 84d8aba Made output directory and output filename more configurable, ...


pom.xml
```diff
@@ -811,5 +732,103 @@
         <debugJavaOptions>-Xdebug -Xrunjdwp:server=y,transport=dt_socket,address=${debugJavaPort},suspend=n</debugJavaOptions>
       </properties>
     </profile>
+    <profile>
```
vjkoskela: The tests are failing because it hasn't started Docker yet. My guess is that the profile is introducing non-determinism in the plugin order. You'll need to rework this as discussed.

ddimensia: Fixed

PR #98, opened 2018-06-01
Metrics Client Host and Port Configuration

vjkoskela:
Allow the metrics client sink to be configured separately from the daemon host and port; however, if not specified fallback to the daemon host and port for backwards compatibility.


BrandonArp:
LGTM. I'm assuming this is useful for containers?

vjkoskela:
We're going to be using it running two MAD instances on a single host. The first will be configured to aggregate a data stream from the clients and the second will be for metrics about metrics. We figured this was easier than running a single MAD with a single source and differentiating the two streams based on tags or prefixes. However, I'd like to talk about how we can make the pipeline configuration more general (we've had similar conversations before).

PR #110, opened 2018-06-29
Remove support for the application v1 protobuf protocol.

vjkoskela:


ddimensia:
I'll need to change our go client before this gets deployed but 👍

BrandonArp:
@ddimensia Keep in mind that this is for the MAD v2  branch.  It's not in master yet.  Though I hope we move to v2 fairly quickly.

PR #111, opened 2018-06-30
Dimension injecting sink.

vjkoskela:


src/main/java/com/arpnetworking/tsdcore/sinks/DimensionInjectingSink.java
```diff
@@ -0,0 +1,111 @@
...
+ */
+public final class DimensionInjectingSink extends BaseSink {
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void recordAggregateData(final PeriodicData data) {
+        final Map<String, String> mergedDimensions = Maps.newHashMap(data.getDimensions().getParameters());
+        mergedDimensions.putAll(_dimensions);
```
ddimensia: Should this force override the dimensions in the data or only inject values for dimensions that don't already exist?  I could see it either way but would probably default to not overriding.

BrandonArp: I agree.  We should make it configurable on whether or not to overwrite and should probably not overwrite by default.

vjkoskela: Done and done. I've separated the configuration into `defaultDimensions` and `overrideDimensions` and inject this into the union map between the data dimensions. Specifically, data overrides defaults and overrides overwrite both.

PR #112, opened 2018-07-02
Allow the configuration to customize the monitoring service.

vjkoskela:
Currently we are running two MAD instances on one of our fleets - clusters - where one instance is used to monitoring and aggregate the local services and the other is used for aggregating remote sample feeds. Since both are on the same cluster we need to customize one of the service names in order to separate out the metrics about metrics.


ddimensia:
👍

PR #133, opened 2018-09-11
Fixes #132 - Add joda dependency to Maven build

### LINKED ISSUE #132: ClassNotFoundException on container startup
The Docker container logs the following error on startup:
```
18:30:46,085 |-ERROR in ch.qos.logback.core.joran.action.NestedComplexPropertyIA - Could not create component [jacksonModule] of type [com.fasterxml.jackson.datatype.joda.JodaModule] java.lang.ClassNotFoundException: com.fasterxml.jackson.datatype.joda.JodaModule
```
The joda dependency appears to not be included in the build.

tmlbl:
Also adds configuration for Maven analyzer so it will not complain that the dependency is unused.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -151,6 +151,24 @@
       </resource>
     </resources>
     <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-dependency-plugin</artifactId>
+        <version>3.1.1</version>
+        <executions>
+          <execution>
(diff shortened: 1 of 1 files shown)


BrandonArp:
It looks like we missed this when we moved from Joda Time to JDK 8 time API.  Instead of adding a module that's not required, lets remove the logger module.

BrandonArp:
Closing in favor of #134

PR #136, opened 2018-09-11
remove joda modules from loggers

BrandonArp:


BrandonArp:
same as #134 but for the 2.0 branch

PR #138, opened 2018-09-17
Remove Http Server Content Length Restriction

vjkoskela:
Lift restrictions on the Akka HTTP server content-length in the default config.


config/config.conf
```diff
@@ -28,5 +28,10 @@ akkaConfiguration {
     stdout-loglevel="DEBUG"
     logging-filter="akka.event.slf4j.Slf4jLoggingFilter"
     actor.debug.unhandled="on"
+    
+    # MAD is effectively a single client service and so a client sending large payments is
```
BrandonArp: Payloads?

PR #145, opened 2019-01-18
Update and resolve dependencies.

vjkoskela:


src/main/java/com/arpnetworking/metrics/mad/parsers/ProtobufV1ToRecordParser.java
```diff
@@ -30,7 +30,6 @@
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
 import com.google.protobuf.InvalidProtocolBufferException;
-import com.inscopemetrics.client.protocol.ClientV1;
```
BrandonArp: Why do you hate imports?

vjkoskela: The class itself is marked deprecated in the latest version of the protocol. The import statement cannot suppress `dreprecation` warnings; the only way I found was to remove the import and use the annotation on the class to suppress the warnings.

PR #149, opened 2019-04-09
Fix settings override

vjkoskela:


.travis.yml
```diff
@@ -1,6 +1,6 @@
 cache:
   directories:
-  - "$HOME/.m2"
+  - "$HOME/.m2"`
```
BrandonArp: I don't think this is right

PR #153, opened 2019-06-21
interfaces and wrappers for kafka source

joeyjackson:


src/main/java/com/arpnetworking/metrics/common/kafka/ConsumerDeserializer.java
```diff
@@ -0,0 +1,64 @@
...
+    public Consumer<K, V> deserialize(final JsonParser parser, final DeserializationContext context)
+            throws IOException, JsonProcessingException {
+        // Parse input json into JsonNode Tree
+        final JsonDeserializer<? extends JsonNode> deserializer = JsonNodeDeserializer.getDeserializer(ObjectNode.class);
+        final JsonNode node = deserializer.deserialize(parser, context);
+
+        // Pull out configs and topics fields and convert deserialize with standard mapper
+        final ObjectMapper mapper = ObjectMapperFactory.getInstance();
+        final TypeReference<Map<String, String>> configsType = new TypeReference<Map<String, String>>() {};
+        final Map<String, Object> configs = mapper.convertValue(node.get("configs"), configsType);
```
speezepearson: nit: It looks like `node.get("configs")` will silently return `null` if `node` is not a JSON object. To avoid this, can we make `node` an `ObjectNode` instead of a generic `JsonNode`?

src/test/java/com/arpnetworking/metrics/common/kafka/ConsumerDeserializerTest.java
```diff
@@ -0,0 +1,143 @@
...
+
+    private static Consumer<String, String> createConsumerFromJSON(final String jsonString, final String fileName)
+            throws IOException {
+        final File file = new File("./target/tmp/filter/ConsumerDeserializer/" + fileName);
+        Files.write(file.toPath(), jsonString.getBytes(Charsets.UTF_8));
+
+
+        final Consumer<String, String> consumer = OBJECT_MAPPER.readValue(
+                new File("./target/tmp/filter/ConsumerDeserializer/" + fileName),
+                new TypeReference<Consumer<String, String>>() {});
```
speezepearson: I believe you can simply use `OBJECT_MAPPER.readValue(jsonString, new TypeReference...)`, and not need to muck about with files!

PR #154, opened 2019-06-29
Kafka source

joeyjackson:


src/test/java/com/arpnetworking/metrics/common/integration/KafkaIT.java
```diff
@@ -0,0 +1,253 @@
...
+                .setPollTimeMillis(POLL_DURATION_MILLIS)
+                .build();
+
+        // Observe records
+        final Observer observer = Mockito.mock(Observer.class);
+        _source.attach(observer);
+        _source.start();
+        for (ProducerRecord<Integer, String> expected : _producerRecords) {
+            Mockito.verify(observer, Mockito.timeout(TIMEOUT)).notify(_source, expected.value());
+        }
```
speezepearson: Out of paranoia, could we verify that those were the _only_ messages processed, and that they were each processed exactly once? I think `Mockito.verify(observer, Mockito.times(_producerRecords.size())).notify(_source, Mockito.any())` would do it.

src/test/java/com/arpnetworking/metrics/common/integration/KafkaIT.java
```diff
@@ -0,0 +1,253 @@
...
+        final KafkaProducer<Integer, String> producer = new KafkaProducer<>(producerProps);
+        producer.initTransactions();
+        try {
+            producer.beginTransaction();
+            for (ProducerRecord<Integer, String> record : records) {
+                producer.send(record);
+            }
+            producer.commitTransaction();
+        } catch (final ProducerFencedException | OutOfOrderSequenceException | AuthorizationException e) {
+            producer.close();
```
speezepearson: It might be appropriate to `Assert.fail()` here? If Kafka gets messed up somehow and doesn't accept the publication, failing here would make the cause of the failure immediately obvious, rather than causing downstream test failures.

(same goes for the next `catch` branch too)

src/main/java/com/arpnetworking/metrics/common/sources/KafkaSource.java
```diff
@@ -0,0 +1,222 @@
...
+                        .log();
+                _runnableConsumer.stop();
+            } else if (throwable instanceof KafkaException) {
+                _logger.error()
+                        .setMessage("Consumer received Kafka Exception")
+                        .addData("source", KafkaSource.this)
+                        .addData("action", "stopping")
+                        .setThrowable(throwable)
+                        .log();
+                _runnableConsumer.stop();
```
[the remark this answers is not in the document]
joeyjackson: Only throws errors for unrecoverable errors

config/pipelines/pipeline.conf
```diff
@@ -123,6 +123,28 @@ sources=[
...
+          bootstrap.servers="localhost:9092"
+          group.id="group0"
+          client.id="consumer0"
+          key.deserializer="org.apache.kafka.common.serialization.StringDeserializer"
+          value.deserializer="org.apache.kafka.common.serialization.StringDeserializer"
+          auto.offset.reset="earliest"
+        }
+      }
+      parser={
+        type="com.arpnetworking.test.StringParser"
```
vjkoskela: Can we configure this by default to use the AMP Protobuf parser?

config/pipelines/pipeline.conf
```diff
@@ -123,6 +123,28 @@ sources=[
...
+        name="StringParser"
+      }
+      pollTimeMillis=1000
```
vjkoskela: We usually express these using ISO8601 durations (e.g. `"PT1S"`). I can point you to an example in the config/code of how to parse these if you like.

pom.xml
```diff
@@ -365,6 +365,26 @@
         <configuration>
           <showLogs>true</showLogs>
           <images>
+            <image>
+              <name>joeyjackson/kafka:${project.version}</name>
```
vjkoskela: Probably shouldn't be `joeyjackson`. I assume we need to do this to avoid a name collision with the one in Dockerhub? 

In that case I would say let's use `arpnetworking/mad/kafka`. 

https://hub.docker.com/r/arpnetworking/mad

However, you need to disable `push` for this image. I believe you can do this by using the `filter` condition in the push execution block above.

https://dmp.fabric8.io/#global-configuration

e.g.
```
          <execution>
            <id>default-push</id>
            <phase>deploy</phase>
            <goals>
              <goal>push</goal>
            </goals>
          </execution>
```

You can add a `<configuration>` block to that which will be merged with the one below. The trick is you need to set the merge policy on `<images>` to not append the execution specific image, but to merge it.

You can find more information about Maven's property management here:

https://blog.sonatype.com/2011/01/maven-how-to-merging-plugin-configuration-in-complex-projects/

pom.xml
```diff
@@ -365,6 +365,26 @@
...
+              <build>
+                <dockerFile>${project.basedir}/src/main/docker/kafka/Dockerfile</dockerFile>
```
vjkoskela: Looks good. Can you also move the MAD `Dockerfile` say into `.../src/main/docker/mad/Dockerfile`? You will need to update the reference in the assembly xml: `src/main/assembly/docker.xml`

src/main/docker/kafka/Dockerfile
```diff
@@ -0,0 +1,30 @@
+# Kafka and Zookeeper
```
vjkoskela: If this is copied from somewhere please provide a reference at the top and include their copyright header. If you changed this substantially, please include our copyright header and a post-copyright comment regarding the original source.

src/main/docker/kafka/scripts/start-kafka.sh
```diff
@@ -0,0 +1,87 @@
+#!/bin/sh
```
vjkoskela: If this is copied from somewhere please provide a reference at the top and include their copyright header. If you changed this substantially, please include our copyright header and a post-copyright comment regarding the original source.

src/main/docker/kafka/supervisor/kafka.conf
```diff
@@ -0,0 +1,4 @@
+[program:kafka]
```
vjkoskela: If this is copied from somewhere please provide a reference at the top and include their copyright header. If you changed this substantially, please include our copyright header and a post-copyright comment regarding the original source.

src/main/docker/kafka/supervisor/zookeeper.conf
```diff
@@ -0,0 +1,4 @@
+[program:zookeeper]
```
vjkoskela: If this is copied from somewhere please provide a reference at the top and include their copyright header. If you changed this substantially, please include our copyright header and a post-copyright comment regarding the original source.

src/main/java/com/arpnetworking/metrics/common/kafka/RunnableConsumerImpl.java
```diff
@@ -0,0 +1,144 @@
...
+import com.arpnetworking.steno.LoggerFactory;
+import net.sf.oval.constraint.NotNull;
+import org.apache.kafka.clients.consumer.Consumer;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.clients.consumer.ConsumerRecords;
+
+import java.time.Duration;
+
+/**
+ * A runnable wrapper for a <code>Consumer</code> that will continually poll
```
vjkoskela: nit: Please use `{@code ...}` or (for in-project references) `{@link ...}` instead of `<code>...</code>`.

src/main/java/com/arpnetworking/metrics/common/kafka/RunnableConsumerImpl.java
```diff
@@ -0,0 +1,144 @@
...
+ * the consumer.
+ *
+ * @param <V> the type of the values in kafka <code>ConsumerRecords</code>
+ *
+ * @author Joey Jackson (jjackson at dropbox dot com)
+ */
+public class RunnableConsumerImpl<V> implements RunnableConsumer {
+    private ConsumerListener<V> _listener;
```
vjkoskela: `final`?

src/main/java/com/arpnetworking/metrics/common/kafka/RunnableConsumerImpl.java
```diff
@@ -0,0 +1,144 @@
...
+    private Consumer<?, V> _consumer;
```
vjkoskela: `final`?

src/main/java/com/arpnetworking/metrics/common/kafka/RunnableConsumerImpl.java
```diff
@@ -0,0 +1,144 @@
...
+    private Duration _pollTime;
```
vjkoskela: `final`?

src/main/java/com/arpnetworking/metrics/common/kafka/RunnableConsumerImpl.java
```diff
@@ -0,0 +1,144 @@
...
+    private volatile boolean _isRunning;
+    private static final Logger LOGGER = LoggerFactory.getLogger(RunnableConsumerImpl.class);
```
vjkoskela: static before non-static; separated by a blank line.

src/main/java/com/arpnetworking/metrics/common/kafka/RunnableConsumerImpl.java
```diff
@@ -0,0 +1,144 @@
...
+                (thread, throwable) -> LOGGER.error()
+                        .setMessage("Unhandled exception")
+                        .setThrowable(throwable)
+                        .log());
+
+        while (isRunning()) {
+            try {
+                final ConsumerRecords<?, V> records = _consumer.poll(_pollTime);
+
+                for (ConsumerRecord<?, V> record : records) {
```
vjkoskela: `record` should be `final`.

src/main/java/com/arpnetworking/metrics/common/kafka/RunnableConsumerImpl.java
```diff
@@ -0,0 +1,144 @@
...
+     * Determine whether the consumer thread is running.
+     *
+     * @return true if running false if not running.
+     */
+    protected boolean isRunning() {
+        return _isRunning;
+    }
+
+    /**
+     * Implementation of builder pattern for <code>RunnableConsumerImpl</code>.
```
vjkoskela: nit: `<code>`; I won't call the rest out.

src/main/java/com/arpnetworking/metrics/common/kafka/RunnableConsumerImpl.java
```diff
@@ -0,0 +1,144 @@
...
+        public RunnableConsumerImpl.Builder<V> setPollTime(final Duration pollTime) {
+            _pollTime = pollTime;
+            return this;
+        }
+
+        @NotNull
+        private Consumer<?, V> _consumer;
+        @NotNull
+        private ConsumerListener<V> _listener;
+        @NotNull
```
vjkoskela: Ideally, you would validate that `_pollTime` is not negative.

src/main/java/com/arpnetworking/metrics/common/kafka/RunnableConsumerImpl.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    /**
+     * Implementation of builder pattern for <code>RunnableConsumerImpl</code>.
+     *
+     * @param <V> the type of the values pulled from kafka records
+     *
+     * @author Joey Jackson (jjackson at dropbox dot com)
+     */
+    public static class Builder<V> extends OvalBuilder<RunnableConsumerImpl<V>> {
```
vjkoskela: nit: `final`

src/main/java/com/arpnetworking/metrics/common/sources/KafkaSource.java
```diff
@@ -0,0 +1,222 @@
...
+ */
+public final class KafkaSource<T, V> extends BaseSource {
+
+    private final Consumer<?, V> _consumer;
+    private final RunnableConsumer _runnableConsumer;
+    private final ExecutorService _consumerExecutor;
+    private final Parser<T, V> _parser;
+    private final Logger _logger;
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(KafkaSource.class);
```
vjkoskela: nit: static before non-static

src/main/java/com/arpnetworking/metrics/common/sources/KafkaSource.java
```diff
@@ -0,0 +1,222 @@
...
+        _consumerExecutor.execute(_runnableConsumer);
+    }
+
+    @Override
+    public void stop() {
+        _runnableConsumer.stop();
+
+        _consumerExecutor.shutdown();
+        try {
+            _consumerExecutor.awaitTermination(10, TimeUnit.SECONDS);
```
vjkoskela: Ideally, the shutdown timeout would be configurable.

vjkoskela: (with a reasonable default)

src/main/java/com/arpnetworking/metrics/common/sources/KafkaSource.java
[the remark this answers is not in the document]
vjkoskela: Can you link the docs here that state that? I would assume an unrecoverable error is typically (only?) encountered on start-up?

Can there be unrecoverable errors that are record specific (since this happens on read)? It seems problematic to kill the consumer in that event.

Specifically, the way the error handler is currently structured, a `ParsingException` will also trigger a stop. Right?

src/main/java/com/arpnetworking/metrics/common/sources/KafkaSource.java
```diff
@@ -0,0 +1,222 @@
...
+
+        @Override
+        protected Builder<T, V> self() {
+            return this;
+        }
+
+        @NotNull
+        private Consumer<?, V> _consumer;
+        @NotNull
+        private Duration _pollTime;
```
vjkoskela: Same comment about duration being non-negative)

src/main/java/com/arpnetworking/metrics/common/sources/KafkaSource.java
```diff
@@ -0,0 +1,222 @@
...
+
+    /**
+     * Builder pattern class for <code>KafkaSource</code>.
+     *
+     * @param <T> the type of data created by the source
+     * @param <V> the type of data of value in kafka <code>ConsumerRecords</code>
+     *
+     * @author Joey Jackson (jjackson at dropbox dot com)
+     */
+    public static class Builder<T, V> extends BaseSource.Builder<Builder<T, V>, KafkaSource<T, V>> {
```
vjkoskela: final

src/test/java/com/arpnetworking/metrics/common/sources/KafkaSourceTest.java
```diff
@@ -0,0 +1,173 @@
...
+            Mockito.verify(observer, Mockito.timeout(TIMEOUT)).notify(_source, expected);
+        }
+        _source.stop();
+    }
+
+    @Test
+    public void testSourceKafkaException() {
+        createExceptionSource(KafkaException.class);
+        _source.start();
+        Mockito.verify(_logBuilder, Mockito.timeout(TIMEOUT)).setMessage("Consumer received Kafka Exception");
```
vjkoskela: Testing via log message is generally a really bad idea. Not only are the tests brittle, they are dependent on the log message actually indicating the behavior/data you want to assert occurred (e.g. a byproduct instead of a direct observation).

config/pipelines/pipeline.conf
joeyjackson: Do you mean change the config file to use that parser or in the case that a parser is not specified, use the protobuf parser (in the source java code)?

vjkoskela: I believe that the code should require a parser since Kafka is a format agnostic source. 

The config you have should use a real parser. I realize this likely means you have to send real data in your tests.

src/main/java/com/arpnetworking/metrics/common/sources/KafkaSource.java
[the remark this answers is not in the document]
joeyjackson: https://kafka.apache.org/22/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#poll-java.time.Duration-

Docs for poll() method list all the exceptions that it throws. All the KafkaExceptions it throws don't seem recoverable except maybe WakeupException which I don't think should ever occur in our use case. These all seem like they would only happen on startup but the consumer doesn't actually "start-up" or connect to the cluster until poll is called.

A Parsing exception would be handled differently because a Kafka ConsumerRecord would be returned so the handle(ConsumerRecord) method would handle a parsing exception by just logging it and dropping that record.

We might not want to stop the consumer for other runtime exceptions caught in the else block but I followed the pattern in FileSource for that case.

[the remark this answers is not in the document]
vjkoskela: = Parsers & Consumers =

Consumer and ConsumerRecord are a Kafka types. Is the current plan to that each parser used by Kafka source implement its own Consumer? 

The way I see it there are two ways to allow re-use of existing parser implementations from Kafka:

1) Implement a Kafka Consumer that can have a configurable Parser. Then attach the Parser to a Kafka source via the standard consumer.
2) Implement a standard consumer that produces a byte-array and configure the Parser on the source.

I'm not sure what the pros and cons of each are. It should be possible to make both equally performant (wrt buffer copies). 

The former has the ability to NAK a request that it cannot parse (I believe) while the latter would make that more complex or impossible. That behavior could also not be varied by parser. I don't think this use case is extremely relevant at this point, as long as we have logs and metrics supporting any failure cases.

However, re-use of the existing parsers for data sourced from Kafka is a key integration requirement.

= Exceptions =

Yes, the FileSource shuts down on an interrupted exception.

https://github.com/ArpNetworking/metrics-aggregator-daemon/blob/master/src/main/java/com/arpnetworking/metrics/common/sources/FileSource.java#L200

However, on all other exceptions it logs, sleeps and continues. The exception may re-occur but it will be logged again and the code backs off. It is more likely to be obvious in the logs should this happen, then for the component to be quietly (it's not silent, but it's not yelling either) and the service to otherwise continue operating normally. 

= Metrics & Logging =

While not necessary on the first pass, I would like you to talk to Spencer about instrumenting the source with scoped or periodic metrics (FYI @speezepearson). This will be critical to successfully deploying and monitoring it.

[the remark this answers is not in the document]
joeyjackson: = Parsers & Consumers =
The current implementation is that the kafka source has a configurable KafkaConsumer and separate configurable Parser. (Case 2 above)
The plan is that the KafkaConsumer would use a standard deserializer to return a ConsumerRecord with a byte[]/bytebuffer value. This can then be passed to the parser to make a mad record. This will allow reuse of the other parsers because most take in a byte array or bytebuffer as input (which kafka already has deserializers for). I dont think there will be any issues with deserializing into a byte[] from kafka so we should always receive a ConsumerRecord. If the parser then cannot parse that array, the kafkasource will just log the parsing error and continue.
The only concern I see with this is that there is currently no validation that the deserialized value from kafka is the same input type as the parser.

= Exceptions =
I dont have strong feelings on this so I changed the handler to sleep and try again when an exception is received (even in "unrecoverable" cases), it will just continually log the error in those cases.

= Metrics & Logging =
TBD

src/test/java/com/arpnetworking/metrics/common/sources/KafkaSourceTest.java
joeyjackson: I agree in general, but in this case isnt logging the actual direct desired behavior in the case of an error (besides the thread sleeping)?

config/pipelines/pipeline.conf
```diff
@@ -123,6 +123,27 @@ sources=[
...
+      name="kafka_source"
+      consumer={
+        type="org.apache.kafka.clients.consumer.Consumer"
+        topics=["topic"]
+        configs={
+          # Set any properties defined at: https://kafka.apache.org/documentation/#consumerconfigs
+          bootstrap.servers="localhost:9092"
+          group.id="group0"
+          client.id="consumer0"
+          key.deserializer="org.apache.kafka.common.serialization.StringDeserializer"
```
vjkoskela: As I understand it, this configures how Kafka reads data out of the message. This doesn't seem appropriate if we're dealing with Protobufs. Wouldn't something that reads into a byte array be safer to avoid mangling due to character set interpretation?

https://kafka.apache.org/10/javadoc/org/apache/kafka/common/serialization/ByteArrayDeserializer.html

src/main/java/com/arpnetworking/metrics/common/sources/KafkaSource.java
```diff
@@ -0,0 +1,287 @@
...
+        super(builder);
+        _logger = logger;
+        _consumer = builder._consumer;
+        _parser = builder._parser;
+        _runnableConsumer = new RunnableConsumerImpl.Builder<V>()
+                .setConsumer(builder._consumer)
+                .setListener(new LogConsumerListener())
+                .setPollTime(builder._pollTime)
+                .build();
+        _consumerExecutor = Executors.newSingleThreadExecutor(runnable -> new Thread(runnable, "KafkaConsumer"));
```
vjkoskela: I think we should make this a fixed thread pool executor with a configurable number of threads.

PR #159, opened 2019-07-22
Kafka concurrent

joeyjackson:


src/test/java/com/arpnetworking/metrics/common/integration/KafkaIT.java
```diff
@@ -189,6 +181,66 @@ public void testKafkaSourceFromConfig() throws IOException {
...
+        final List<String> expected = _producerRecords.stream()
+                .map(ProducerRecord::value)
+                .collect(Collectors.toList());
+        Collections.sort(expected);
+        Collections.sort(collected);
+        Assert.assertEquals(expected, collected);
+    }
+
+    @Test
+    public void testKafkaSourceFillQueue() throws InterruptedException {
```
speezepearson: If the entire purpose of this test is to make sure that the `KafkaSource` still works correctly when the queue fills up, it doesn't seem like it needs to be an integration test -- can we turn it into a unit test?

It would also be nice -- though possibly too hard to be worth it -- to enforce, in this test, that the queue does indeed fill up. I think you would have to inject some sort of synchronization primitive into the KafkaSource (maybe an optional Semaphore that throttles the consumers, or a custom BlockingQueue implementation that you could use to wait until `put` has been called `queueSize+1` times before letting any `get` complete). Testing threading is hard, though, and it might turn out to be more work than it's worth.

src/test/java/com/arpnetworking/metrics/common/integration/KafkaIT.java
```diff
@@ -189,6 +181,66 @@ public void testKafkaSourceFromConfig() throws IOException {
...
+        _source = new KafkaSource.Builder<String, String>()
+                .setName("KafkaSource")
+                .setParser(new StringParser())
+                .setConsumer(_consumer)
+                .setPollTime(POLL_DURATION)
+                .setNumWorkerThreads(4)
+                .build();
+
+        // Observe records
+        final CollectObserver observer = new CollectObserver();
```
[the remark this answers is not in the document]
joeyjackson: I like this method better

src/main/java/com/arpnetworking/metrics/common/sources/KafkaSource.java
```diff
@@ -138,15 +138,37 @@ private KafkaSource(final Builder<T, V> builder) {
         _parserExecutor = Executors.newFixedThreadPool(_numWorkerThreads);
         _shutdownAwaitTime = builder._shutdownAwaitTime;
         _backoffTime = builder._backoffTime;
+        _bufferSize = builder._bufferSize;
+        _buffer = new ArrayBlockingQueue<>(_bufferSize);
+    }
+
+    // NOTE: Package private for testing
+    /* package private */ KafkaSource(final Builder<T, V> builder, final BlockingQueue<V> buffer) {
```
speezepearson: [nit] We have these two ~identical package-private constructors; could we refactor out the commonality into another constructor that they simply delegate to?

PR #160, opened 2019-07-24
Kafka instrument

joeyjackson:


src/test/java/com/arpnetworking/metrics/common/sources/KafkaSourceTest.java
```diff
@@ -99,102 +124,208 @@ public void testSourceMultiWorkerSuccess() {
...
+                .setPeriodicMetrics(_periodicMetrics),
                 _logger);
     }
 
+    private void assertMetricCountsCorrect(final long inRecords, final long outRecords, final long parsingExceptions,
+                                           final long kafkaExceptions, final long consumerExceptions) {
+        _periodicMetrics.run(); // Flush the metrics
+        final List<Long> expectedInRecords =
+                _periodicMetrics.getCounters(_source._recordsInCountMetricName) == null
+                ? Collections.singletonList(0L)
```
cwbriones: I would remove the burden of these checks by documenting/annotating that `CollectorPeriodicMetrics#getCounters` always returns a non-null list and just return a `Collections.emptyList()` by default. 

This way all of these ternary expressions can go away since the sum of an empty list is still 0.

src/test/java/com/arpnetworking/test/CollectorPeriodicMetrics.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    /**
+     * Returns the list of collected counters for a specific metric.
+     *
+     * @param name the name of the metric
+     * @return the list of counters
+     */
+    public List<Long> getCounters(final String name) {
+        return _counts.get(name);
```
cwbriones: See my above comment in the test, but you should return _counts.getOrDefault with an empty list here to avoid having to handle the null case.

src/test/java/com/arpnetworking/test/CollectorPeriodicMetrics.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    /**
+     * Returns the list of collected gauges for a specific metric.
+     *
+     * @param name the name of the metric
+     * @return the list of gauges
+     */
+    public List<Number> getGauges(final String name) {
```
cwbriones: remove this and getTimers since they're unused by the tests.

PR #166, opened 2019-09-11
Telemetry Integration Test

vjkoskela:


src/test/java/com/arpnetworking/test/TelemetryClient.java
```diff
@@ -0,0 +1,276 @@
...
+        final AsyncHttpClient asyncHttpClient = Dsl.asyncHttpClient();
+        _webSocket = asyncHttpClient.prepareGet(String.format("ws://%s:%d/telemetry/v2/stream", host, port))
+                .execute(
+                        new WebSocketUpgradeHandler.Builder()
+                                .addWebSocketListener(new TelemetryWebSocketListener(host, port, _callbacks))
+                                .build())
+                .get();
+
+        _keepAliveExecutor = Executors.newScheduledThreadPool(1);
+        _keepAliveExecutor.schedule(
```
speezepearson: Does this action want to be recurring? It looks to me like it will only run once.

src/test/java/com/arpnetworking/test/TelemetryClient.java
```diff
@@ -0,0 +1,276 @@
...
+            final Consumer<Double> callback) {
+        try {
+            _webSocket.sendTextFrame(
+                    OBJECT_MAPPER.writeValueAsString(
+                            ImmutableMap.of(
+                                    "command", "subscribeMetric",
+                                    "service", service,
+                                    "metric", metricName,
+                                    "statistic", statistic.getName())));
+            _callbacks.put(String.format("%s/%s/%s", service, metricName, statistic.getName()), callback);
```
speezepearson: Consider refactoring `String.format("%s/%s/%s", service, metricName, statisticName)` into a static method to force consistency?

src/test/java/com/arpnetworking/test/TelemetryClient.java
vjkoskela: Good catch! Fixed.

src/test/java/com/arpnetworking/test/TelemetryClient.java
vjkoskela: Done!

PR #176, opened 2020-04-29
Update DimensionInjectingSink to use ThreadLocalBuilder for PeriodicData

vjkoskela:


vjkoskela:
Yes, but more useful to us is that it removes a warning from the log files. Less clutter should hopefully make it clearer when something real is broken/misbehaving.

PR #178, opened 2020-04-29
Concurrent modification exception

vjkoskela:


src/main/java/com/arpnetworking/metrics/mad/PeriodWorker.java
```diff
@@ -182,11 +182,10 @@ private void processRecord(final Record record) {
         int closedBucketCount = 0;
 
         // Phase 1: Collect expired buckets
-        for (final ZonedDateTime key : expiredBucketMap.keySet()) {
-            for (final Bucket bucket : _bucketsByExpiration.remove(key)) {
-                expiredBuckets.add(bucket);
-            }
+        for (final Map.Entry<ZonedDateTime, List<Bucket>> entry : expiredBucketMap.entrySet()) {
```
orborde, line 185: It seems like this change is trying to fix a ConcurrentModificationException that results from multiple threads executing the changed block concurrently.

Assuming I've got that right: it seems like this could still throw a ConcurrentModificationException if another thread calls `expiredBucketMap.clear()` on its view of the underlying `NavigableMap` while this iteration is happening.

Am I missing something?

vjkoskela: Yep, there is no concurrent thread access to this since the PeriodWorker was converted to an actor.

The concurrent access resulted from me optimizing away the use of ConcurrentMap as the backing stores (since it was now single threaded), but not looking over the logic carefully enough to realize that if two times are closed at once (rare) that it will attempt to modify the map inside an iteration.

This is not valid for standard maps in Java; only ConcurrentMaps (even by a single thread). This rearranges the logic to not read the map after modification.

orborde: Oh, the CME was because `_bucketsByExpiration.remove()` was touching the same underlying map that `expiredBucketMap` pointed to? It seems like that problem should have been present even before you changed it to remove the possibility of concurrent thread access.

But, yeah, regardless, this change does seem to be switching this code from an incorrect state to a correct one.

PR #179, opened 2020-04-29
Normalize (and imo improve) periodic statistics collection in MAD.

vjkoskela:
* Adds mappedDimension to PeriodicStatisticsSink to allow renaming of dimension extracted from the data stream and partitioning the output periodic statistics; (e.g. `_namespace` to `namespace`)
* Converts identifier `String` creation to use `Key` for metric+dimension-key (aka statistic tracking).
* Removes statistic from that tracking; MAD installs now only compute histogram internally.
* Removes period from that tracking; it doesn't really make sense (imo).
* Simplifies the `AggregatedData` iteration.
* Clarifies the usage of PeriodicStatistics sink; namely, that it should be used in a single period sink-chain.

### DIFF
--- a/src/main/java/com/arpnetworking/tsdcore/sinks/PeriodicStatisticsSink.java
+++ b/src/main/java/com/arpnetworking/tsdcore/sinks/PeriodicStatisticsSink.java
@@ -390,9 +390,9 @@ public final class PeriodicStatisticsSink extends BaseSink {
         private final LongAccumulator _metricSamples = new LongAccumulator(Long::sum, 0);
         private final AtomicLong _aggregatedData = new AtomicLong(0);
         private final AtomicReference<Set<String>> _uniqueMetrics = new AtomicReference<>(
-                Collections.newSetFromMap(Maps.newConcurrentMap()));
+                ConcurrentHashMap.newKeySet());
         private final AtomicReference<Set<MetricKey>> _uniqueStatistics = new AtomicReference<>(
-                Collections.newSetFromMap(Maps.newConcurrentMap()));
+                ConcurrentHashMap.newKeySet());
     }
 
     /**
--- a/src/main/java/com/arpnetworking/tsdcore/sinks/PeriodicStatisticsSink.java
+++ b/src/main/java/com/arpnetworking/tsdcore/sinks/PeriodicStatisticsSink.java
@@ -35,8 +35,8 @@ import net.sf.oval.constraint.CheckWith;
 import net.sf.oval.constraint.CheckWithCheck;
 import net.sf.oval.constraint.Min;
 import net.sf.oval.constraint.NotNull;
-import net.sf.oval.constraint.ValidateWithMethod;
 
+import java.time.Duration;
 import java.util.Collections;
 import java.util.Deque;
 import java.util.Iterator;
@@ -51,6 +51,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.atomic.LongAccumulator;
+import javax.annotation.Nullable;
 
 /**
  * Aggregates and periodically logs metrics about the aggregated data being
@@ -63,13 +64,6 @@ import java.util.concurrent.atomic.LongAccumulator;
  * More recently, this class has been extended to allow attribution of usage
  * by bucketing and tagging the emitted periodic self-instrumentation metrics
  * using configured tag keys who's values are lighted from the data flow itself.
- *
- * IMPORTANT: This sink counts total samples through it and will count samples
- * across periods. It is strongly recommended that you configure this sink on
- * a sink-chain for a single period (e.g. using PeriodFilteringSink). Further,
- * it assumes that all AggregatedData instances for a given metric contain a
- * correct PopulationSize and are thus interchangeable as far as sample counting
- * is concerned.
  *
  * This class is thread safe.
  *
@@ -89,7 +83,11 @@ public final class PeriodicStatisticsSink extends BaseSink {
 
         // Record the periodic data against the appropriate bucket
         // NOTE: if no dimensions are configured then the bucket key is an empty map
-        final Key bucketKey = computeKey(periodicData.getDimensions(), _mappedDimensions);
+        final Key bucketKey = computeKey(
+                periodicData.getDimensions(),
+                _periodDimensionName,
+                periodicData.getPeriod(),
+                _mappedDimensions);
         _buckets.computeIfAbsent(bucketKey, k -> new Bucket(bucketKey))
                 .record(periodicData, now);
     }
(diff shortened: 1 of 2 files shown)


src/main/java/com/arpnetworking/tsdcore/sinks/PeriodicStatisticsSink.java
```diff
@@ -401,11 +435,30 @@ protected Builder self() {
             return this;
         }
 
+        private boolean validate(final ImmutableMap<String, String> ignored) {
```
cwbriones, line 438: `ignored` corresponds to `_mappedDimensions` in this case right?

Can you just rename the parameter to `mappedDimensions` and use that?

src/main/java/com/arpnetworking/tsdcore/sinks/PeriodicStatisticsSink.java
```diff
@@ -240,47 +247,55 @@ public void run() {
...
-                if (uniqueMetrics.add(serviceMetric)) {
-                    // Allow each service/metric in the periodic data to contribute
-                    // its population size (samples processed) to the sample count.
-                    _metricSamples.accumulate(datum.getPopulationSize());
-                }
+            final ImmutableMultimap<String, AggregatedData> data = periodicData.getData();
+            _aggregatedData.addAndGet(data.size());
+
+            for (final String metricName : data.keySet()) {
+                final AggregatedData firstDatum = data.get(metricName).iterator().next();
```
cwbriones, line 288: I'm confused here how this is equivalent to the above. In the older code, we were iterating over all metricName, AggregatedData entries but here we're only taking the first one we see for a given metricname.

For `_uniqueMetrics` and `_uniqueStatistics` this would be equivalent sure, since the other AggregatedData would have the same set of dimensions but wouldn't `_metricSamples` be thrown off?

src/main/java/com/arpnetworking/tsdcore/sinks/PeriodicStatisticsSink.java
```diff
@@ -105,6 +113,24 @@ public Object toLogValue() {
                 .build();
     }
 
+    static Key computeKey(final Key key, final ImmutableMultimap<String, String> mappedDimensions) {
```
cwbriones, line 116: Maybe make a note about how this "flattens" the mappedDimensions and only keeps the last one?

src/main/java/com/arpnetworking/tsdcore/sinks/PeriodicStatisticsSink.java
```diff
@@ -401,11 +435,30 @@ protected Builder self() {
...
 
+        private boolean validate(final ImmutableMap<String, String> ignored) {
+            final Set<String> mappedTargetDimensions = Sets.newHashSet(_mappedDimensions.values());
+
+            // Assert that no two mapped dimensions target the same value
+            if (mappedTargetDimensions.size() != _mappedDimensions.size()) {
+                return false;
+            }
+
+            // Assert that no mapped dimension target is also a dimension
```
cwbriones, line 446: Would it be correct to rename a dimension and subsequently replace it?

e.g.
```
dimensions = {"a", "b"}
mappedDimensions = {"b": "c", "a" : "b"}
```
We replace `b` with `c`, and then `a` is mapped to `b` in the output. Because input `b` was itself mapped, there is no conflict. I admit that this is a bit contrived but it is a situation that doesn't have a conflict so I would expect it to work.

src/main/java/com/arpnetworking/tsdcore/sinks/PeriodicStatisticsSink.java
cwbriones: Ah nevermind this comment, I see that the javadoc says that this should only take in periodicData for a single period, so this list would only have one element.

I think I would still rather see that precondition asserted instead of silently emitting incorrect metrics.

src/main/java/com/arpnetworking/tsdcore/sinks/PeriodicStatisticsSink.java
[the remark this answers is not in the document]
vjkoskela: I would if this were a single field validation. 

Unfortunately, in this case we are validating `_dimensions` against `_mappedDimensions` and by convention I just reference both members. 

I've converted it to a class level `@CheckWith`.

src/main/java/com/arpnetworking/tsdcore/sinks/PeriodicStatisticsSink.java
```diff
@@ -384,6 +402,22 @@ public Builder setDimensions(final ImmutableSet<String> value) {
             return this;
         }
 
+        /**
+         * Incoming dimension names to partition the periodic statistics on
+         * mapped to an outbound dimension name. Cannot be null. Default is
+         * empty map (no dimension partitions).
+         *
+         * Note that if the same dimension is specified here and in
+         * dimensions the output key from mapped dimensions takes precedence.
```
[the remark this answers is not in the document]
cwbriones: I see. So we only dedupe on output dimensions, but using the same input dimension and aliasing into two or more outputs is legal.

Makes sense. Thanks for the clarification.

src/main/java/com/arpnetworking/tsdcore/sinks/PeriodicStatisticsSink.java
```diff
@@ -147,28 +173,28 @@ private void flushMetrics() {
         }
     }
 
-    private static Set<String> createConcurrentSet(final Set<String> existingSet) {
+    private static <T> Set<T> createConcurrentSet(final Set<T> existingSet) {
```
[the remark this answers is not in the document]
vjkoskela: Not sure how that applies. The contents of the `existingSet` are not important here, rather making sure that the new concurrent set is scaled to handle approximately the same volume of data.

src/main/java/com/arpnetworking/tsdcore/sinks/PeriodicStatisticsSink.java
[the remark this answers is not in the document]
vjkoskela: Correct. That allows you to (for whatever reason) duplicate an input dimension.

src/main/java/com/arpnetworking/tsdcore/sinks/PeriodicStatisticsSink.java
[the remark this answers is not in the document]
cwbriones: Probably should have left this on the `return Collections.newSetFromMap` line below.

src/main/java/com/arpnetworking/tsdcore/sinks/PeriodicStatisticsSink.java
vjkoskela: The example you provided is invalid because `b` is mapped to `b` (via `dimensions) and `a` is also mapped to `b` (with value from `a`; via `mappedDimensions`) so there is an output conflict here.

I describe a case below where you have no effectively conflict; say if your example were:
```
dimensions = {"a", "b"}
mappedDimensions = {"b": "c", "b" : "b"}
```

Here b is in both lists as a target (detected as a conflict) but the effect is a no-op since it would end up having the same value (input `b`) in both cases.

vjkoskela:
Not at the moment. The better solution in my opinion is to support validated with method on the class level.

PR #181, opened 2020-05-04
Upgrade histogram bucket counts to 64-bit.

vjkoskela:
Will not succeed until this is released:

https://github.com/ArpNetworking/metrics-aggregator-protocol/pull/10

### DIFF
--- a/src/main/java/com/arpnetworking/metrics/mad/model/statistics/HistogramStatistic.java
+++ b/src/main/java/com/arpnetworking/metrics/mad/model/statistics/HistogramStatistic.java
@@ -20,9 +20,9 @@ import com.arpnetworking.metrics.mad.model.DefaultQuantity;
 import com.arpnetworking.metrics.mad.model.Quantity;
 import com.arpnetworking.metrics.mad.model.Unit;
 import com.arpnetworking.tsdcore.model.CalculatedValue;
-import it.unimi.dsi.fastutil.doubles.Double2IntAVLTreeMap;
-import it.unimi.dsi.fastutil.doubles.Double2IntMap;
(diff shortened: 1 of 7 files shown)


src/main/java/com/arpnetworking/metrics/mad/model/statistics/HistogramStatistic.java
```diff
@@ -375,9 +375,10 @@ public Double getValueAtPercentile(final double percentile) {
             // The Math.min is for the case where the computation may be just
             // slightly larger than the _entriesCount and prevents an index out of range.
             final int target = (int) Math.min(Math.ceil(_entriesCount * percentile / 100.0D), _entriesCount);
-            int accumulated = 0;
-            for (final Double2IntMap.Entry next : _data.double2IntEntrySet()) {
-                accumulated += next.getIntValue();
+            // TODO(ville): Is it sufficient for this to be a long?
```
speezepearson, line 378: `accumulated` will be bounded by the count of the histogram, right? So if the count can be a `long`, this can too?

vjkoskela: Yes, `accumulated` here and `count` in the augmented-histogram both represent the total number of samples represented. Both should be `long`.  

My question here is more forward looking that if individual bucket counts can overflow an `int` is `accumulated` (and `count`) sufficiently large as a `long`? Will we ever have histograms with more than 9,223,372,036,854,775,808 samples? 

=> I think not, it would be 42M samples per minute per client into a daily roll-up over 150M clients.

PR #186, opened 2020-05-13
Periodic Statistics Default Dimension Values

vjkoskela:
Add support to periodic statistics to use default values for partitioning dimensions.


src/main/java/com/arpnetworking/tsdcore/sinks/PeriodicStatisticsSink.java
```diff
@@ -116,7 +117,8 @@ static Key computeKey(
             final Key key,
             final String periodDimensionName,
             final Duration period,
-            final ImmutableMultimap<String, String> mappedDimensions) {
+            final ImmutableMultimap<String, String> mappedDimensions,
```
cwbriones, line 120: Looking at this code a second time, this parameter makes **much** more sense when I think of it as "dimension key aliases" and forget that it came from mapping them at all (since that's not relevant to the method body)

PR #188, opened 2020-05-16
Track the minimum request timestamp aggregated into a MAD bucket, and push that to CAGG during flush

orborde:
This PR is incomplete. Sending for early feedback to check that:
* I've found all the code points that I need to touch.
* I'm writing idiomatic Java for this project.

Obviously I plan to add tests. And, you know, get it to actually compile (regrettably, this depends on https://github.com/ArpNetworking/metrics-aggregator-daemon/pull/187 and https://github.com/ArpNetworking/metrics-aggregator-protocol/pull/13).


vjkoskela:
FYI. The failure of this pull request to fail is a bug in the `jdk-wrapper.sh` script. Unfortunately my testing on Mac turned out to be inadequate and the background job management fails on Linux (Ubuntu 18.x) while it passes on Mac. I am converting the script to a simpler execution model and will release tomorrow and patch ArpNetworking projects accordingly.

vjkoskela:
Here's the `jdk-wrapper` fix in question:

https://github.com/KoskiLabs/jdk-wrapper/pull/44

src/main/java/com/arpnetworking/tsdcore/model/PeriodicData.java
```diff
@@ -42,6 +43,8 @@ public ZonedDateTime getStart() {
         return _start;
     }
 
+    public ZonedDateTime getMinRequestTime() { return _min_request_time; }
```
vjkoskela, line 46: nit: we generally avoid single line functions even for accessors.

src/main/java/com/arpnetworking/metrics/mad/model/Record.java
```diff
@@ -43,6 +43,13 @@
      */
     ZonedDateTime getTime();
 
+    /**
+     * Gets the "received at" time stamp of the record.
+     *
+     * @return the time stamp
+     */
+    ZonedDateTime getRequestTime();
```
vjkoskela, line 51: This should be `Optional<ZonedDateTime>`.

src/main/java/com/arpnetworking/metrics/mad/model/DefaultRecord.java
```diff
@@ -198,6 +216,8 @@ protected void reset() {
         @NotNull
         private ZonedDateTime _time;
         @NotNull
```
vjkoskela, line 218: This field should be `@Nullable` right since it's not on all requests?

src/main/java/com/arpnetworking/metrics/mad/model/DefaultRecord.java
```diff
@@ -45,6 +45,11 @@ public ZonedDateTime getTime() {
         return _time;
     }
 
+    @Override
+    public ZonedDateTime getRequestTime() {
```
vjkoskela, line 49: This should be `Optional<ZonedDateTime>`.

src/main/java/com/arpnetworking/metrics/mad/Bucket.java
```diff
@@ -178,6 +180,16 @@ public void add(final Record record) {
             }
             addMetric(metric, calculators);
         }
+
+        if (_min_request_time == null) {
+            _min_request_time= record.getRequestTime();
+        }
+
+        if (record.getRequestTime() != null) {
+            if (record.getRequestTime().isBefore(_min_request_time)) {
```
vjkoskela, line 189: Does `isBefore` thrown on `null`?

src/main/java/com/arpnetworking/metrics/mad/Bucket.java
```diff
@@ -349,6 +361,7 @@ private void addMetric(
     }
 
     private boolean _isOpen = true;
+    private ZonedDateTime _min_request_time = null;
```
vjkoskela, line 364: Prefer `Optional` to `null`, or as you are mutating it at least mark it `@Nullable`.

vjkoskela:
Perhaps you could elaborate on:

> I'm writing idiomatic Java for this project.

The project, for historical reasons, has adopted some styles that are not those endorsed by Sun/Oracle. However, the code is moving in that direction. Most of the style (not confident to say all) is rather self-evident in the code and even enforced by Checkstyle.

In general, the plumbing looks good. Although it will be worth verifying; manual end-to-end test seems like the easiest way, perhaps having CAGG log the value it received.

It does look like you did not update the TCP `AggregationServerSink` which should receive the same treatment as its HTTP counterpart.

src/main/java/com/arpnetworking/metrics/mad/Bucket.java
orborde: Refactored!

src/main/java/com/arpnetworking/metrics/mad/Bucket.java
orborde: Made it `Optional`! Should I have made `DefaultRecord._request_time` an `Optional` as well?

It's not clear to me when I should use `null` vs. `Optional`.

orborde:
> Perhaps you could elaborate on:
> 
> > I'm writing idiomatic Java for this project.

I don't have any specifics I'm confused about (beyond when to use `Optional` vs `@Nullable`). I simply figured that, since I have heretofore spent almost no time rolling patches for Java, I would almost certainly get some of the idioms wrong and need someone to point them out in review.

orborde:
> It does look like you did not update the TCP `AggregationServerSink` which should receive the same treatment as its HTTP counterpart.

Done!

src/main/java/com/arpnetworking/metrics/mad/model/DefaultRecord.java
orborde: Why @Nullable instead of `Optional<ZonedDateTime>`?

vjkoskela: Builder setters are used to deserialize from formats like JSON and HOCON and this must accept `null`. It was easier to have these setters always accept objects (e.g. `Double` not `double`) and deal with `null` on validation then to use `Optional` and get the parsers to do the conversion correctly in all cases because we definitely don't want to deal with an `Optional` that is either set, unset or `null`.

src/main/java/com/arpnetworking/metrics/mad/model/DefaultRecord.java
```diff
@@ -100,13 +107,15 @@ private DefaultRecord(final Builder builder) {
...
         _time = builder._time;
+        _request_time = Optional.ofNullable(builder._request_time);
         _annotations = builder._annotations;
         _dimensions = builder._dimensions;
     }
 
     private final ImmutableMap<String, ? extends Metric> _metrics;
     private final String _id;
     private final ZonedDateTime _time;
+    private final ZonedDateTime _request_time;
```
vjkoskela, line 118: We typically make the member `Optional` to ensure any internal logic is also forced to deal with its optionality as well as to avoid creating a new `Optional` on every access.

src/main/java/com/arpnetworking/tsdcore/model/PeriodicData.java
```diff
@@ -90,12 +95,14 @@ public String toString() {
...
         _period = builder._period;
         _start = builder._start;
+        _min_request_time=builder._min_request_time;
         _dimensions = builder._dimensions;
         _data = builder._data;
     }
 
     private final Duration _period;
     private final ZonedDateTime _start;
+    private final ZonedDateTime _min_request_time;
```
vjkoskela, line 105: This should be `Optional` too, right?

src/main/java/com/arpnetworking/metrics/mad/parsers/ProtobufV3ToRecordParser.java
```diff
@@ -70,6 +70,7 @@
                         DefaultRecord.Builder.class,
                         builder -> builder.setId(new UUID(high, low).toString())
                                 .setTime(ZonedDateTime.ofInstant(Instant.ofEpochMilli(record.getEndMillisSinceEpoch()), ZoneOffset.UTC))
+                                .setRequestTime(ZonedDateTime.ofInstant(Instant.ofEpochMilli(record.getRequestMillisSinceEpoch()), ZoneOffset.UTC))
```
vjkoskela, line 73: This probably should not be called if the record does not have a `requestMillisSinceEpoch`.

src/main/java/com/arpnetworking/tsdcore/sinks/AggregationServerHttpSink.java
```diff
@@ -76,6 +76,7 @@ protected String getContentType() {
                 .setMetric(metricName)
                 .setPeriod(periodicData.getPeriod().toString())
                 .setPeriodStart(periodicData.getStart().toString())
+                .setMinRequestTime(periodicData.getMinRequestTime().toString())
```
vjkoskela, line 79: You will then need to deal with the Optionality of the field here and in the `AggregationServerSink` as well; maybe use `ifPresent`.

src/main/java/com/arpnetworking/metrics/mad/parsers/ProtobufV3ToRecordParser.java
orborde: Seems likethe way this code is organized means that I'd have to copypaste the whole Builder block if I wanted to add an if statement. What do you think of this?

```
.setRequestTime(maybeParseMillisSinceEpoch(record.getRequestMillisSinceEpoch()).orElse(null))
```

vjkoskela: No, that should not be necessary, just convert the single line lambda into a method; e.g.

```
records.add(ThreadLocalBuilder.build(
    DefaultRecord.Builder.class,
    builder -> {
           builder.setId(new UUID(high, low).toString())
               .setTime(ZonedDateTime.ofInstant(
                   Instant.ofEpochMilli(record.getEndMillisSinceEpoch()), ZoneOffset.UTC))
               .setDimensions(buildDimensions(record))
               .setMetrics(buildMetrics(record))
           if (record.getRequestMillisSinceEpoch() != ) {
               builder.setRequestTime(...);
           }
           return builder;
    }));
```

The above would negate the need for the `maybeParseMillisSinceEpoch` method entirely, since the conditional allows you to encapsulate that code too and that logic is not currently reused. 

However, your method works too!

vjkoskela: General note; in Java there are three kinds of lambdas:

1) Method reference; e.g. MyClass::doFoo
2) Single expression; e.g. builder -> builder.setFoo(...)
3) Inline method; e.g. builder -> { builder.setFoo(...); if (x) { builder.setBar(...) }; return builder }

Oracle reference here:

https://docs.oracle.com/javase/tutorial/java/javaOO/lambdaexpressions.html#syntax

(not a terrific read imo; better off if you're interested in picking up Effective Java v3).

src/main/java/com/arpnetworking/metrics/mad/Bucket.java
```diff
@@ -349,6 +361,7 @@ private void addMetric(
     }
 
     private boolean _isOpen = true;
+    private Optional<ZonedDateTime> _min_request_time;
```
vjkoskela, line 364: Camel case; `_minRequestTime`

src/main/java/com/arpnetworking/metrics/mad/model/DefaultRecord.java
```diff
@@ -100,13 +107,15 @@ private DefaultRecord(final Builder builder) {
...
         _time = builder._time;
+        _request_time = Optional.ofNullable(builder._request_time);
         _annotations = builder._annotations;
         _dimensions = builder._dimensions;
     }
 
     private final ImmutableMap<String, ? extends Metric> _metrics;
     private final String _id;
     private final ZonedDateTime _time;
+    private final Optional<ZonedDateTime> _request_time;
```
vjkoskela, line 118: Camel case; `_requestTime`

PS. The `_` prefix is used to distinguish members from locals and arguments and avoid name collisions.

src/main/java/com/arpnetworking/metrics/mad/model/DefaultRecord.java
```diff
@@ -197,6 +217,8 @@ protected void reset() {
         private String _id;
         @NotNull
         private ZonedDateTime _time;
+        @Nullable
+        private ZonedDateTime _request_time;
```
vjkoskela, line 221: Camel case; _requestTime

src/main/java/com/arpnetworking/tsdcore/model/PeriodicData.java
```diff
@@ -90,12 +96,14 @@ public String toString() {
...
         _period = builder._period;
         _start = builder._start;
+        _min_request_time = Optional.ofNullable(builder._min_request_time);
         _dimensions = builder._dimensions;
         _data = builder._data;
     }
 
     private final Duration _period;
     private final ZonedDateTime _start;
+    private final Optional<ZonedDateTime> _min_request_time;
```
vjkoskela, line 106: Camel case; `_min_request_time`

src/main/java/com/arpnetworking/tsdcore/model/PeriodicData.java
```diff
@@ -167,6 +187,9 @@ protected void reset() {
         private Duration _period;
         @NotNull
         private ZonedDateTime _start;
+        @Nullable
+        private ZonedDateTime _min_request_time;
```
vjkoskela, line 191: Camel case; _min_request_time

src/main/java/com/arpnetworking/metrics/mad/parsers/ProtobufV3ToRecordParser.java
orborde: ugh, I wanted to do the *clever* *functional* Optional-using thing, but honestly your suggested lambda is much more concise and easier to read.

PR #191, opened 2020-05-19
Backport histogram fix from CAGG port.

vjkoskela:


cwbriones:
Can you provide context here by either including a description or linking to the relevant MAD diff?

Happy to stamp afterwards.

vjkoskela:
The CAGG diff in question was: 

https://github.com/ArpNetworking/metrics-cluster-aggregator/pull/108

And the change is in response to @speezepearson reviewing that change which ported the code without this change to CAGG. This delta is feedback addressed when the code was merged to CAGG and now being addressed in MAD (the origin of the CAGG code).

PR #204, opened 2020-06-04
Widen PeriodicDataToProtoConverter's responsibilities to subsume a little more of the repeated code at its callers, and add a unit test

orborde:


src/main/java/com/arpnetworking/tsdcore/model/PeriodicDataToProtoConverter.java
```diff
@@ -38,43 +37,25 @@
...
-     */
-    public PeriodicDataToProtoConverter(final PeriodicData periodicData) {
-        _period = periodicData.getPeriod();
-        _periodStart = periodicData.getStart();
-        _dimensionParameters = periodicData.getDimensions().getParameters();
-        _cluster = periodicData.getDimensions().getCluster();
-        _service = periodicData.getDimensions().getService();
-    }
-
     /**
```
vjkoskela, line 40: nit: Javadoc not required on private methods

src/main/java/com/arpnetworking/tsdcore/model/PeriodicDataToProtoConverter.java
```diff
@@ -103,6 +84,31 @@ public PeriodicDataToProtoConverter(final PeriodicData periodicData) {
...
+
+    }
+
+    /**
+     * Convert a PeriodicData to a set of corresponding protobuf messages.
+     *
+     * @param periodicData PeriodicData being converted.
+     * @return List of StatisticSetRecord protobufs corresponding to the above.
+     */
+    public static List<Messages.StatisticSetRecord> convert(
```
vjkoskela, line 97: nit: Typically public before private in method ordering within class

src/main/java/com/arpnetworking/tsdcore/model/PeriodicDataToProtoConverter.java
```diff
@@ -134,4 +140,8 @@ private static ByteString serializeSupportingData(final AggregatedData datum) {
         }
         return byteString;
     }
+
+    private PeriodicDataToProtoConverter() {
+        throw new AssertionError("can't construct utility class");
```
vjkoskela, line 145: You don't need to throw in here. Just marking it private is enough to prevent someone from instantiating it.

PR #206, opened 2020-06-09
Emit eviction metric

abbeyqy:


src/main/java/com/arpnetworking/tsdcore/sinks/HttpPostSinkActor.java
```diff
@@ -200,6 +201,9 @@ private void processEmitAggregation(final EmitAggregation emitMessage) {
             }
 
             if (evicted > 0) {
+                final Metrics metrics = _sink.getMetricsFactory().create();
```
vjkoskela, line 204: You should use try-with-resources to ensure it's closed even on exception.

src/main/java/com/arpnetworking/tsdcore/sinks/HttpPostSink.java
```diff
@@ -176,17 +158,13 @@ protected HttpPostSink(final Builder<?, ?> builder) {
         _aysncHttpClientUri = Uri.create(_uri.toString());
 
         _sinkActor = builder._actorSystem.actorOf(
-                HttpPostSinkActor.props(CLIENT, this, builder._maximumConcurrency, builder._maximumQueueSize, builder._spreadPeriod));
-        _metricsFactory = builder._metricsFactory;
-
-        _evictedRequestsName = "sinks/http_post/" + getMetricSafeName() + "/evicted_requests";
+                HttpPostSinkActor.props(CLIENT, this, builder._maximumConcurrency, builder._maximumQueueSize,
```
vjkoskela, line 161: nit: convention in these projects is that once you line break, each argument goes on its own line.

e.g.
```
_sinkActor = builder._actorSystem.actorOf(
    HttpPostSinkActor.props(
        CLIENT, 
        this, 
        builder._maximumConcurrency, 
        builder._maximumQueueSize,
        builder._spreadPeriod, 
        builder._metricsFactory));
```

PR #207, opened 2020-06-11
Fix docker assembly by ensuring the directory is clean first.

vjkoskela:


vjkoskela:
Requires: https://github.com/ArpNetworking/arpnetworking-parent-pom/pull/60

PR #215, opened 2020-07-30
Add samples_dropped metric to MAD

abbeyqy:


abbeyqy:
`computePopulationSize` method in `AggregationServerHttpSink` refers to the code here: https://github.com/ArpNetworking/metrics-cluster-aggregator/commit/e40d6dc9f88341a961b99b1953db6b07be55f301#diff-eba1035dfa98d0f4ef8e755ac5ce7f35R121
and `deserialize` method in `AggregationMessage` is almost identical to https://github.com/ArpNetworking/metrics-cluster-aggregator/blob/master/src/main/java/com/arpnetworking/tsdcore/model/AggregationMessage.java#L58

src/main/java/com/arpnetworking/tsdcore/model/AggregationMessage.java
```diff
@@ -103,6 +109,87 @@ public ByteString serializeToByteString() {
...
 
+
+    /**
+     * Deserialize message from {@link Buffer}.
+     *
+     * @param data The {@link Buffer} containing the serialized message.
+     * @return The deserialized {@link AggregationMessage} or absent if
+     * the {@link Buffer} could not be deserialized.
+     */
+    public static Optional<AggregationMessage> deserialize(final ByteString data) {
```
wittekm, line 120: Is there really no utility that would do this for you?! wild

src/main/java/com/arpnetworking/tsdcore/sinks/AggregationServerHttpSink.java
```diff
@@ -47,20 +57,78 @@ protected String getContentType() {
...
+            serializedPeriodicData.add(new SerializedDatum(
+                    AggregationMessage.create(record).serializeToByteString().toArray(),
+                    computePopulationSize(record.getMetric(), record.getStatisticsList())
+                    )
+            );
         }
         return serializedPeriodicData.build();
     }
 
+    private static long computePopulationSize(final String metricName, final List<Messages.StatisticRecord> statisticsList) {
```
vjkoskela, line 74: So I dont think this is how we want to do it. Based on our previous conversation the assumption was that we could obtain this information from the Protobuf structures. However, it's clear that this is not possible because we have to deserialize the supporting data. 

I think it would be better to push the logic down into `PeriodicDataToProtoConverter` and have it return a structure which contained the population size with each `Messages.StatisticSetRecord`.

src/main/java/com/arpnetworking/tsdcore/model/AggregationMessage.java
abbeyqy: No... we write this function in CAGG so I basically copied that part from CAGG.

src/main/java/com/arpnetworking/tsdcore/model/AggregationMessage.java
```diff
@@ -103,6 +109,87 @@ public ByteString serializeToByteString() {
...
+                            LOGGER.warn(
+                                    String.format("Invalid protocol buffer, unknown subtype; type=%s, subtype=%s, bytes=%s",
+                                            type,
+                                            subType,
+                                            Hex.encodeHexString(payloadBytes)));
+                            return Optional.empty();
+                    }
+                default:
+                    LOGGER.warn(String.format("Unsupported message type; type=%s", type));
+                    return Optional.empty();
```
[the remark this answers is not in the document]
wittekm: doesnt need to happen though

PR #219, opened 2020-08-05
Enable configuration of monitoring sinks.

vjkoskela:


config/config.conf
```diff
@@ -13,8 +13,29 @@ httpPort=7090
...
+#    dispatchErrorLoggingInterval = "PT1M"
+#    unsupportedDataLoggingInterval = "PT1M"
+#  }
+#]
 #jvmMetricsCollectionInterval="PT1.0S"
 
+# NOTES:
+# - metricsClientHost and metricsClientPort are deprecated
+# - metricsClientHost and metricsClientPort no longer default from httpHost and httpPort
+# - you CANNOT specify either these AND monitoringSinks
```
cwbriones, line 35: This wording is a bit awkward, specifically the "either and" construction.

Perhaps something more like "If either metricsClientHost or metricsClientPort are specified, then the legacy configuration will take precedence over monitoringSinks"

src/main/java/com/arpnetworking/metrics/mad/configuration/AggregatorConfiguration.java
```diff
@@ -337,6 +375,14 @@ public Builder setAkkaConfiguration(final Map<String, ?> value) {
         @NotEmpty
         private String _monitoringService = "mad";
         @NotNull
+        private ImmutableList<JsonNode> _monitoringSinks;
+        @NotEmpty
+        private String _metricsClientHost;
```
cwbriones, line 380: Should this and `metricsClientPort` be marked as `@Nullable` now? The `ofNullable` above isn't correct upon inspection without this

src/main/java/com/arpnetworking/metrics/mad/Main.java
```diff
@@ -302,6 +316,36 @@ private Injector launchGuice(final ActorSystem actorSystem) {
...
+        final List<com.arpnetworking.metrics.Sink> sinks = new ArrayList<>();
+        for (final JsonNode sinkNode : monitoringSinks) {
+            @Nullable final JsonNode classNode = sinkNode.get("class");
+            try {
+                if (classNode != null) {
+                    final Class<?> builderClass = Class.forName(classNode.textValue() + "$Builder");
+                    final Object builder = OBJECT_MAPPER.treeToValue(sinkNode, builderClass);
+                    @SuppressWarnings("unchecked")
+                    final com.arpnetworking.metrics.Sink sink =
+                            (com.arpnetworking.metrics.Sink) builderClass.getMethod("build").invoke(builder);
```
cwbriones, line 337: 🤢 

I get why this hack is necessary, it's just really ugly.

src/main/java/com/arpnetworking/metrics/mad/parsers/CollectdJsonToRecordParser.java
```diff
@@ -280,7 +281,7 @@ private CollectdRecord(final Builder builder) {
          * {@link com.arpnetworking.commons.builder.Builder} implementation for
          * {@link CollectdRecord}.
          */
-        public static final class Builder extends ThreadLocalBuilder<CollectdRecord> {
+        public static final class Builder extends OvalBuilder<CollectdRecord> {
```
cwbriones, line 284: unclear why this changed

src/main/java/com/arpnetworking/metrics/mad/parsers/GraphitePlaintextToRecordParser.java
```diff
@@ -219,7 +220,7 @@ private GraphitePlaintextToRecordParser(final Builder builder) {
      * {@link com.arpnetworking.commons.builder.Builder} implementation for
      * {@link GraphitePlaintextToRecordParser}.
      */
-    public static final class Builder extends ThreadLocalBuilder<GraphitePlaintextToRecordParser> {
+    public static final class Builder extends OvalBuilder<GraphitePlaintextToRecordParser> {
```
cwbriones, line 223: unclear why this changed

src/main/java/com/arpnetworking/metrics/mad/parsers/JsonToRecordParser.java
```diff
@@ -714,7 +715,7 @@ private static Quantity version2gSampleToQuantity(@Nullable final Version2g.Samp
      * {@link com.arpnetworking.commons.builder.Builder} implementation for
      * {@link JsonToRecordParser}.
      */
-    public static final class Builder extends ThreadLocalBuilder<JsonToRecordParser> {
+    public static final class Builder extends OvalBuilder<JsonToRecordParser> {
```
cwbriones, line 718: again, unclear why this changed

src/main/java/com/arpnetworking/metrics/mad/Main.java
vjkoskela: I started with the preferred approach, but there were some nasty surprises in the metrics client. I put it on hold, and hope to return to it in the future to help undo this hack.

src/main/java/com/arpnetworking/metrics/mad/parsers/CollectdJsonToRecordParser.java
vjkoskela: The parser is instantiated in a way that does not use the ThreadLocalBuilder correctly resulting in warnings in the log. It's not related to this PR; just discovered in the testing of this PR.

In general, TLB should be used in areas where objection creation velocity is high. That is not the case with Parsers or Sinks. Therefore, in these cases it is sufficient to use OB.

config/config.conf
vjkoskela: Done

src/main/java/com/arpnetworking/metrics/mad/configuration/AggregatorConfiguration.java
vjkoskela: Done.

src/main/java/com/arpnetworking/metrics/mad/parsers/CollectdJsonToRecordParser.java
cwbriones: Makes sense, thanks for clarifiying!

BrandonArp: I think you changed the wrong one to OvalBuilder here.  I don't think the CollectdJsonToRecordParser class has a builder.

PR #220, opened 2020-08-14
Convert HttpPostSinkActor Request class Builder to OvalBuilder from TLB.

vjkoskela:


vjkoskela:
Issue to address: https://github.com/ArpNetworking/metrics-aggregator-daemon/issues/221

BrandonArp:
Why does TLB not work?

vjkoskela:
The TLB setters are called from two different classes on the same instance of the TLB. This effectively disables its functionality as a TLB. The code needs to be refactored to make ownership of the TLB instance clearer or support a callback style population of the TLB.

PR #225, opened 2020-10-15
Record failure count in http post sink

joeyjackson:
Add a metric to HttpPostSink to record count of post request failures (successes are already counted).

This allows for the rate of success to be calculated instead of just raw numbers


cwbriones:
> This allows for the rate of success to be calculated instead of just raw numbers

Do you mean failure rate? We can already calculate the per-route success rate as `avg(sinks/http_post/<route>/successes)`

joeyjackson:
You are correct. I misunderstood the metric so this PR is unnecessary

PR #226, opened 2020-10-22
Use time construct for RequestEntry enterTime 

joeyjackson:
In the RequestEntry and HttpSink queue latency, change usage to time.Instant for type and time.Duration for arithmetic on timestamps instead of long for semantic clarity


vjkoskela (approved):
Pre-req; build passing.

PR #232, opened 2021-01-18
add jenkinsfile

BrandonArp:
Moving to our own build infrastructure. With the loss of Travis, I was finally incentivized to setup a Jenkins instance. This should let us build more advanced scenarios as well as use more resources in our build (and get fancy plugins to report). Project admins should contact me to setup an account on Jenkins to control the builds.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -35,7 +35,7 @@ pipeline {
             usernamePassword(credentialsId: 'jenkins-ossrh', usernameVariable: 'OSSRH_USER', passwordVariable: 'OSSRH_PASS'),
             string(credentialsId: 'jenkins-gpg', variable: 'GPG_PASS')]) {
           withMaven {
-            sh "./jdk-wrapper.sh ./mvnw $target"
+            sh "./jdk-wrapper.sh ./mvnw clean $target -P rpm -U -B -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn"
           }
         }
       }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,51 @@
+pipeline {
+  agent {
+    kubernetes {
+      yamlFile 'kubes-pod.yaml'
+      defaultContainer 'ubuntu'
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,51 @@
...
+        script {
+          target = "install"
+        }
+      }
+    }
+    stage('Setup release') {
+      when { buildingTag(); not { changeRequest() }  }
+      steps {
+        script {
+          target = "install deploy -P release  --settings settings.xml"
```
vjkoskela, line 28: Shouldn't this have `-P rpm` as well?

vjkoskela: There are some other useful flags we've used in the past as well:

* `-U` - Update all package versions (avoids negative caching)
* `-B` - Batch mode short circuits any prompts for user input
* `-Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn` - Reduces logging

vjkoskela: Ref:
https://github.com/ArpNetworking/metrics-aggregator-daemon/blob/master/.github/workflows/build.yml#L32

Jenkinsfile
```diff
@@ -0,0 +1,51 @@
...
+    stage('Init') {
+      steps {
+        checkout scm
+      }
+    }
+    stage('Setup build') {
+      when { not { buildingTag() } }
+      steps {
+        script {
+          target = "install"
```
vjkoskela, line 20: Missing Maven arguments? (see below)

vjkoskela (changes requested):

Jenkinsfile
BrandonArp: Updated. What is the stuff from today about the pooled connections and keep-alives?

vjkoskela: It wasn't a problem on Travis, and I would be surprised it was a problem egressing your build network. I believe you do pay a penalty wrt performance with the settings that were prescribed for Azure. I'm okay going back to baseline here.

vjkoskela (approved):

PR #236, opened 2021-01-22
Emit a metric of samples received by the aggregator.

vjkoskela:


src/main/java/com/arpnetworking/metrics/mad/Aggregator.java
```diff
@@ -122,6 +125,33 @@ public void notify(final Observable observable, final Object event) {
...
+                        .map(m -> {
+                            final ImmutableList<CalculatedValue<?>> countStatistic = m.getStatistics().get(
+                                    STATISTIC_FACTORY.getStatistic("count"));
+                            if (countStatistic != null) {
+                                return countStatistic.stream()
+                                        .map(s -> s.getValue().getValue())
+                                        .reduce(Double::sum)
+                                        .orElse(0.0d)
+                                        +
+                                        m.getValues().size();
```
cwbriones, line 141: Just to sanity check - you're summing both the count (if available) and the number of values because the protocol allows for both aggregate statistics and raw samples, correct?

vjkoskela: Yes, that's correct.

cwbriones (approved):

PR #238, opened 2021-02-16
Actor reap

vjkoskela:


src/main/java/com/arpnetworking/metrics/mad/PeriodWorker.java
```diff
@@ -92,40 +109,57 @@ public void preRestart(final Throwable reason, final Optional<Object> message) t
...
-            Duration timeToRotate = Duration.between(now, rotateAt.get());
+            Duration timeToRotate = Duration.between(now, _nextScheduledRotationTime.get());
             if (timeToRotate.isNegative()) {
                 timeToRotate = Duration.ZERO;
             } else if (timeToRotate.compareTo(MINIMUM_ROTATION_CHECK_INTERVAL) < 0) {
                 timeToRotate = MINIMUM_ROTATION_CHECK_INTERVAL;
             }
 
-            _nextScheduledRotation = context().system().scheduler().scheduleOnce(
+            _nextScheduledRotationTask = context().system().scheduler().scheduleOnce(
```
BrandonArp, line 157: Please use the timers() call to schedule a single-shot timer. This will also eliminate/simplify some of the clean-up code as the base class will handle canceling named timers.

BrandonArp (changes requested):

src/main/java/com/arpnetworking/metrics/mad/PeriodWorker.java
vjkoskela: Done

src/main/java/com/arpnetworking/metrics/mad/PeriodWorker.java
```diff
@@ -16,70 +16,83 @@
...
     }
 
     @Override
     public void postStop() {
-        if (_nextScheduledRotation != null) {
-            final boolean cancelResult = _nextScheduledRotation.cancel();
-            _nextScheduledRotation = null;
+        _periodicMetrics.recordCounter("actors/period_worker/stopped", 1);
+
+        timers().cancel(IDLE_CHECK_TIMER);
```
BrandonArp, line 83: You don't actually need to do this. But I think you should call super.postStop().

BrandonArp (approved):

src/main/java/com/arpnetworking/metrics/mad/PeriodWorker.java
vjkoskela: Added super calls.

src/main/java/com/arpnetworking/metrics/mad/Aggregator.java
```diff
@@ -342,6 +461,17 @@ public Builder setPeriods(final Set<Duration> value) {
             return this;
         }
 
+        /**
+         * Set the idle timeout for workers. Cannot be null.
```
cwbriones, line 465: Maybe add a comment here about the actor reaping occurring when the timeout is reached.

src/main/java/com/arpnetworking/metrics/mad/Aggregator.java
```diff
@@ -279,12 +228,182 @@ private Aggregator(final Builder builder) {
...
     private final LoadingCache<String, Optional<ImmutableSet<Statistic>>> _cachedSpecifiedStatistics;
     private final LoadingCache<String, Optional<ImmutableSet<Statistic>>> _cachedDependentStatistics;
-    private final Map<Key, List<ActorRef>> _periodWorkerActors = Maps.newConcurrentMap();
 
     private static final StatisticFactory STATISTIC_FACTORY = new StatisticFactory();
-    private static final Timeout SHUTDOWN_TIMEOUT = Timeout.apply(1, TimeUnit.SECONDS);
+    private static final Duration SHUTDOWN_TIMEOUT = Duration.ofSeconds(30);
     private static final Logger LOGGER = LoggerFactory.getLogger(Aggregator.class);
 
+    private static final class ShutdownAggregator implements Serializable {
```
cwbriones, line 236: Why is this message type a singleton instead of a string?

I noticed that string types are used for singleton messages in the other actor.

cwbriones (commented):
I didn't see the new idle behavior documented anywhere, not picky about where you place it but I think it should at least be in one of the builder docstrings.

LGTM overall. The semantics of the idle timeout are slightly different than in CAGG – the way akka does it with the cluster sharding is a timeout from the time of the last received record message. To make this behavior the same you would have to reset the idle check timer every time you receive a new record. 

I don't think this is necessarily an issue though so long as the idle timeout is greater than the aggregation window. I think in this specific case it's handled by not stopping the actor if there's a rotation scheduled.

src/main/java/com/arpnetworking/metrics/mad/Aggregator.java
vjkoskela: Done!

src/main/java/com/arpnetworking/metrics/mad/Aggregator.java
vjkoskela: Generally, string instances are okay for internal messages. I had this originally also being sent from Aggregator to PeriodWorker (just like it does in CAGG). However, it was changed to just a PoisonPill. I suspect in the future we will be sending ShutdownAggregator to the PeriodWorker to allow it to kill itself gracefully (including flushing of state to disk).

vjkoskela:
Yes, the semantics are somewhat different. The idle timeout is an interval, and one full interval must elapsed with no messages received. Therefore, reaping does not occur exactly after X minutes of idle, but rather between X and 2X. This avoids the expense of resetting the timer on every received message. If we want to reduce the delta we can have a shorter regular interval and it can check the time delta between the timestamp since last message and current time. This would reduce the drift to just X to X+T where T is the time interval between checks.

cwbriones:
> Therefore, reaping does not occur exactly after X minutes of idle, but rather between X and 2X. If we want to reduce the delta we can have a shorter regular interval and it can check the time delta between the timestamp since last message and current time. 

Makes sense. I'm fine with this I just thought to point out the difference. It's simpler this way too which is a plus, and really the important part is performing the actor shutdown at all.

PR #239, opened 2021-02-19
Fix Maven on GHA by overriding the http wagon with a custom build.

vjkoskela:


vjkoskela:
See:
* https://issues.apache.org/jira/browse/FLINK-16947
* https://issues.apache.org/jira/browse/WAGON-607
* https://issues.apache.org/jira/browse/WAGON-608

.github/fix-maven.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+MAVEN_HOME=$(./jdk-wrapper.sh ./mvnw -v |grep 'Maven home:' | awk '{ print $3 }')
```
cwbriones, line 2: nit. awk already matches lines 🙂  `./jdk-wrapper.sh ./mvnw -v | awk '/^Maven home:/ { print $3 }' `

.github/fix-maven.sh
```diff
@@ -0,0 +1,11 @@
...
+if [ -d "$MAVEN_HOME" ]; then
+  echo "Installing custom wagon-http jar to $MAVEN_HOME/lib"
+  ls -l "$MAVEN_HOME/lib/"wagon-http-*-shaded.jar
+  rm "$MAVEN_HOME/lib/"wagon-http-*-shaded.jar
+  cp ./.github/wagon-http-* "$MAVEN_HOME/lib/"
```
cwbriones, line 7: Because this is in the .m2 directory it will be included in the cache in subsequent runs. We may want to include this file in the cache key so that if we need to update the build it doesn't place two copies of wagon into this directory (one from cache, one from the hypothetical update).

.github/fix-maven.sh
vjkoskela: Yeah, I just cribbed it and didn't try to make it better; just validated it works. Fixed!

FYI. This will all be deleted once we are able to release from the private Jenkins instance.

.github/fix-maven.sh
vjkoskela: That's a great point.  I don't anticipate going back to that before we can move to Jenkins, but it's a good idea. Unfortunately, since the build has already run as part of the PR the JAR is already there.

https://github.com/actions/cache/issues/2

So if we change the key now, removing it later will revert to a cache that has the problem. I think I'm going to punt addressing this until that happens and we can change the cache key at that point. 

Added a note to the build file.

cwbriones: Sounds good, I'm fine with punting so long as we're aware of it.

cwbriones (approved):

PR #242, opened 2021-03-24
Fix Actor Reaping Bottleneck

vjkoskela:
The underlying cause of the change in memory profile was a growing akka actor queue backlog. This was verified through elevated age self-instrumentation metrics.

The problem lies in the conversion of a multi-threaded observer to a single threaded actor instance. 

The solution is to have a pool of actors each responsible for a partition of period workers. Since period workers are keyed by dimension set, the hash of the set was used as the partition modulo 2*cores (the number of aggregators).

The correctness was verified through head-to-head testing against version 1.18.0 (pre actor reaping) and against version 1.19.2 (with broken actor reaping). System memory, JVM memory, JVM GC metrics as well as freshness and stability all score on-par with version 1.18.0 after this change and dramatically better than version 1.19.2. 

Additionally, temporary per-aggregator metrics were added to verify reasonable distribution via the hash code among actors. Obviously, this is workload dependent.

### DIFF
--- a/src/main/java/com/arpnetworking/metrics/mad/Aggregator.java
+++ b/src/main/java/com/arpnetworking/metrics/mad/Aggregator.java
@@ -58,11 +58,9 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
-import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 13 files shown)


src/main/java/com/arpnetworking/metrics/mad/PeriodWorker.java
```diff
@@ -305,6 +305,10 @@ private void processRecord(final Record record) {
     private final Bucket.Builder _bucketBuilder;
     private final NavigableMap<ZonedDateTime, Bucket> _bucketsByStart = new TreeMap<>();
     private final NavigableMap<ZonedDateTime, List<Bucket>> _bucketsByExpiration = new TreeMap<>();
+
+    // WARNING: Consider carefully the volume of samples recorded.
+    // PeriodicMetrics reduces the number of scopes creates, but each sample is
+    // still stored in-memory until it is flushed.
```
cwbriones, line 311: You may want to add this to the docstring on the metrics client itself, since this advice still goes outside of this project.

src/main/java/com/arpnetworking/metrics/mad/Aggregator.java
```diff
@@ -129,8 +146,12 @@ public void notify(final Observable observable, final Object event) {
             return;
         }
 
+        // TODO(ville): Should the record contain a key instead of just annotations?
+        // ^ This raises the bigger question of metric name as part of the key.
+        // (at the moment it's not taking advantage of same key-space across metrics in bucket)
         final Record record = (Record) event;
-        _actor.tell(record, ActorRef.noSender());
+        final int actorIndex = (record.getDimensions().hashCode() & 0x7FFFFFFF) % _actors.size();
```
cwbriones, line 153: hashCode isn't particularly uniform for partitioning purposes. Perhaps MD5 here instead?

More of a nit, the modulo will bias messages to the actors at the front of the set. That's probably a very small issue here given the (relatively) small number of actors.

src/main/java/com/arpnetworking/metrics/mad/Aggregator.java
```diff
@@ -271,124 +341,121 @@ public Key getKey() {
...
+
+                // Report any failures
+                final boolean success = periodWorkerShutdown.stream()
+                        .map(f -> f.getNow(false))
+                        .reduce(Boolean::logicalAnd)
+                        .orElse(true);
+                if (!success) {
+                    LOGGER.error()
+                            .setMessage("Failed stopping one or more period worker actors")
+                            .addData("aggregatorActor", this.self())
```
cwbriones, line 393: nit. a few places in this file use `self()` and others use `this.self()`. Can we standardize on the former?

cwbriones (commented):
I'd prefer if the non-behavioral refactor (message handlers into methods) was separate in the future. It would make it easier to spot the meaningful changes.

---
The core of this code looks correct in terms of logic but I don't think I fully understand the problem this addresses.

> The underlying cause of the change in memory profile was a growing akka actor queue backlog

queue as in actor message queue?

If that's so, then my read is that the aggregator throughput was not high enough when there was a single actor per `Aggregation`. Switching to a partitioned pool of actors increases throughput enough to avoid the message backlog from building and thus prevents excessive memory usage.

I am a bit surprised that this is CPU-bound vs IO-bound. I suppose reaggregating histograms is more expensive than I expected.

src/main/java/com/arpnetworking/metrics/mad/PeriodWorker.java
vjkoskela: The periodic metrics class is experimental at the moment. This has been one of the learnings I've been discussing with @BrandonArp and as we fold it into the core library it will be addressed. However, in the meantime we are using the experimental functionality in MAD and CAGG to facilitate instrumentation.

cwbriones (approved):

cwbriones:
Approved, we synced on this and my understanding was correct.

src/main/java/com/arpnetworking/metrics/mad/Aggregator.java
vjkoskela: Yes, I totally agree.

The flip side is that computing the MD5 hash of the dimensions will result in additional overhead.

The skew from the modulo is pretty minimal as a percent of the keyspace. 

I will add a comment regarding these issues.

src/main/java/com/arpnetworking/metrics/mad/Aggregator.java
vjkoskela: Done!

Issue #481, opened 2024-08-30
Histogram precision is not propagated from MAD to CAgg and KairosDB


BrandonArp:
Each step assumes 7 bits. Need a field added to the proto, parsing in cagg, and then support in the cagg sink.

PR #482, opened 2024-08-31
Proto precision

BrandonArp:


BrandonArp:
fixes #481 (along with #arpnetworking/metrics-cluster-aggregator#407)

